PR #7, opened 2026-04-21
perf: store Decimal directly in elaboration::Amount

alevy:
## Summary

- Changes `elaboration::Amount` from `BTreeMap<Commodity, [u8; 16]>` to `BTreeMap<Commodity, Decimal>`
- Removes all `Decimal::serialize()`/`Decimal::deserialize()` calls from the elaboration hot path
- Updates `src/main.rs` and `benches/queries.rs` to use `Decimal` directly
- Adds manual `Serialize`/`Deserialize` impls for `Amount` that preserve the `[u8; 16]` wire format, so existing `.bki` files remain readable

## Motivation

Every posting amount previously required a round-trip through `[u8; 16]`: amounts were serialized to bytes for storage in the map, then deserialized back to `Decimal` for every arithmetic operation. This happened on every posting in every transaction during elaboration. Storing `Decimal` directly eliminates this overhead entirely.

The `.bki` binary format is **unchanged** — the custom serde impls on `Amount` convert `Decimal ↔ [u8; 16]` during serialization so the on-disk representation is byte-for-byte identical to the original.

## Benchmark results

All workloads use 100k transactions. Results are median times.

| Benchmark | Before | After | Δ |
|-----------|--------|-------|---|
| elaboration/simple | 9.82 ms | 9.61 ms | −2.1% |
| elaboration/wide | 12.65 ms | 12.32 ms | −2.6% |
| elaboration/deep | 24.02 ms | 23.91 ms | −0.5% |
| elaboration/multi_commodity | 10.46 ms | 10.52 ms | +0.6% |
| compile/simple | 65.95 ms | 67.40 ms | +2.2% |
| compile/wide | 69.61 ms | 71.18 ms | +2.3% |
| compile/deep | 181.65 ms | 185.76 ms | +2.3% |
| compile/multi_commodity | 107.36 ms | 105.98 ms | −1.3% |

The elaboration stage itself shows a consistent improvement. The compile-level numbers are within noise — the dominant cost at that level is parsing (~65 ms of the 67 ms total), so the elaboration saving is not visible end-to-end. The code simplification is the primary benefit.

## Test plan

- [x] `cargo test --lib` — 20 tests pass (including 2 new: wire-format compat and round-trip)
- [x] Benchmarks run before and after on identical workloads

### DIFF
--- a/benches/queries.rs
+++ b/benches/queries.rs
@@ -26,7 +26,7 @@ fn bench_balance(c: &mut Criterion) {
                                 .entry(posting.account.as_str())
                                 .or_default()
                                 .entry(commodity.as_str())
-                                .or_default() += Decimal::deserialize(*amount);
+                                .or_default() += *amount;
                         }
                     }
                 }
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -128,7 +128,7 @@ pub type Commodity = String;
 /// via postcard without needing a custom codec, while preserving exact decimal
 /// precision (no floating-point rounding).
 #[derive(Deserialize, Default, Serialize, Debug)]
-pub struct Amount(pub BTreeMap<Commodity, [u8; 16]>);
+pub struct Amount(pub BTreeMap<Commodity, Decimal>);
 
 /// Cleared/pending state of a resolved transaction or posting.
 ///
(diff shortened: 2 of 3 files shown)


alevy:
Can we refactor the `Amount` type to still serialize (perhaps with serde's `with` directive?) using Decimal's `serialize` and `deserialize` under the hood while preserving the `Decimal` type once in Rust?

alevy:
Done — replaced the derive with manual `Serialize`/`Deserialize` impls that convert `Decimal ↔ [u8; 16]` during serde calls. The `.bki` wire format is now byte-for-byte identical to the original, while the Rust API exposes `Decimal` directly. Added two tests: one that asserts the postcard bytes match a raw `BTreeMap<_, [u8;16]>` (wire compat), and one full round-trip test.

PR #46, opened 2026-04-21
Add accounts, commodities, and stats subcommands

### LINKED ISSUE #20: Add accounts, commodities, and stats subcommands
Three straightforward CLI subcommands not yet implemented:

**`accounts`** — List all account names that appear in the journal, optionally filtered by a pattern. `elaboration::Journal` already tracks accounts; this is mostly plumbing in `src/main.rs`.

**`commodities`** — List all commodity symbols seen in the journal. Iterate `journal.transactions`, collect commodity keys from each posting's `Amount` map, deduplicate.

**`stats`** — Summary statistics: number of transactions, date range (first/last transaction date), number of unique accounts, number of commodities. All computable from `elaboration::Journal` with no new data structures.

**Opinion:** All three are low-effort and high-value for ledger-cli compatibility. `stats` is especially useful for quick sanity checks on a journal file.

alevy:
Closes #20

## Summary

- **`accounts`**: lists all accounts that appear in the journal (from `journal.accounts`, a sorted `BTreeMap`), one per line. Accepts an optional case-insensitive substring `PATTERN` to filter results.
- **`commodities`**: collects all commodity symbols used across every posting's `amount` map, deduplicates and sorts them via `BTreeSet`, then prints one per line.
- **`stats`**: prints a compact summary table with transaction count, unique account count, unique commodity count, and the first/last transaction date (decoded from CE-epoch days via `chrono::NaiveDate::from_num_days_from_ce_opt`).

## Test plan

- [ ] `cargo build` passes cleanly
- [ ] `cargo test --lib` — all 33 tests pass
- [ ] `cargo clippy --bin ledger` — no new warnings introduced by this change

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, fs::File, io::{Read as _, Write as _}, path::PathBuf};
+use std::{collections::{BTreeMap, BTreeSet}, fs::File, io::{Read as _, Write as _}, path::PathBuf};
 
 use clap::{Parser, Subcommand};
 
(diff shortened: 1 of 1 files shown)


alevy:
Fixed a bug in the `stats` subcommand before merging: `from_num_days_from_ce_opt` interprets its argument as days since 0001-01-01 (CE epoch), but `ResolvedTransaction.date` is documented as days since 1970-01-01 (Unix epoch). The fix uses the same Unix epoch arithmetic as the `register` subcommand.

PR #51, opened 2026-04-21
Wire HIR::prices through to elaboration::Journal

### LINKED ISSUE #14: Wire HIR::prices through to elaboration::Journal
PR #9 added `HIR::prices: Vec<HistoricalPrice>` populated from `P` directives during resolution. However, `elaboration::Journal` never reads this field — historical prices are dropped during elaboration and not accessible to callers.

**Fact:** `HIR::prices` is populated but discarded. See resolution.rs (HIR struct `prices` field) and elaboration.rs (Journal struct — no `prices` field).

**To fix:** Add `prices: Vec<HistoricalPrice>` (or equivalent) to `elaboration::Journal` and populate it from `HIR::prices` during `Journal::try_from`. This will expose historical prices to callers for market-value queries.

Related: CLI issue "Add `prices` subcommand" (part of issue #16 or similar).

alevy:
## Summary

- Adds `elaboration::HistoricalPrice`, a fully evaluated price record with `date` (epoch-days `i32`), `time`, `commodity`, `price` (`Decimal`), and `price_commodity`
- Adds `Journal::prices: Vec<HistoricalPrice>` so callers can read historical price quotes from the elaborated journal
- `TryFrom<HIR> for Journal` now evaluates each `P` directive's price expression using the final context and populates the field; previously these were silently dropped

Closes #14

## Test plan

- [ ] New test `elaboration::tests::test_prices_wired_through_to_journal` constructs an `ast::Journal` with one `P` directive, runs it through `resolution::HIR::try_from` then `elaboration::Journal::try_from`, and asserts `journal.prices` contains the expected evaluated price entry
- [ ] `cargo test --lib` — all 34 tests pass

### DIFF
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -46,6 +46,24 @@ pub struct Journal {
     pub transactions: Vec<ResolvedTransaction>,
     /// All accounts referenced by any posting, with their declared properties.
     pub accounts: BTreeMap<String, AccountProperties>,
+    /// Market price quotes from `P` directives, in source order, with the
+    /// price expression fully evaluated to a concrete `(Decimal, commodity)`.
+    pub prices: Vec<HistoricalPrice>,
+}
+
+/// A fully evaluated historical price entry produced from a `P` directive.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct HistoricalPrice {
+    /// Days since the Unix epoch on which this price was recorded.
+    pub date: i32,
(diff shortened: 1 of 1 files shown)


alevy:
One minor note before merge: prices are evaluated using the *last* context in the file rather than the context at their source position. This means a `define` alias appearing after a `P` directive would still apply to that price's expression. Unlikely to matter in practice (price expressions are almost always literal amounts), but worth tracking.

Issue #30, opened 2026-04-21
Add write_ledger convenience function


alevy:
Part of #18. Add a `write_ledger<W: Write>(w: W, txns: &[resolution::Transaction])` function (probably in `src/lib.rs`) that serialises a slice of transactions to any `Write`-implementing sink using the existing `Display` impls. This makes it easy to generate `.ledger` files programmatically. Depends on #18a (Transaction builder) being useful first.

alevy:
Implemented in PR #59 — `write_ledger<W: Write>(txns, writer)` added to `src/lib.rs`.

PR #59, opened 2026-04-21
feat: add write_ledger<W: Write>() library function (#30)

alevy:
## Summary

- Implements `write_ledger<W: Write>(entries: impl IntoIterator<Item=resolution::Transaction>, writer: &mut W) -> io::Result<()>` in `src/lib.rs`, closing issue #30
- Exports the function from the crate root so callers use `ledger::write_ledger(...)`
- Wires the existing `Print` subcommand to call `write_ledger()` instead of its own inline loop, keeping library and CLI in sync

## Design

- Transactions are formatted via their existing `Display` impl and separated by blank lines
- The function accepts any `W: Write` sink (file, `Vec<u8>`, `BufWriter`, stdout, etc.)
- No trailing blank line is emitted after the last transaction

## Tests (6 new, all in `src/lib.rs`)

- `write_empty_iterator_produces_no_output` — empty input yields empty output
- `write_single_transaction_basic` — correct text, no trailing blank line
- `multiple_transactions_separated_by_blank_line` — blank-line separator, no trailing double newline
- `round_trip_preserves_date_and_description` — date, description, state, postings survive parse → resolve
- `round_trip_preserves_metadata_and_tags` — tags, comments, and key-value metadata survive round-trip
- `round_trip_multiple_transactions` — two transactions both survive round-trip correctly
- `round_trip_posting_with_metadata` — per-posting metadata and tags survive round-trip

## Test plan

- [x] `cargo test --lib` — 51/51 pass
- [x] `cargo test` (including doctests) — 54/54 pass
- [x] `cargo build` — clean build, no new warnings from changed code

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,16 +169,7 @@ mod write_ledger_tests {
         };
         let ast_journal = p.parse(&source.to_string()).expect("parse failed");
         let hir: resolution::HIR = ast_journal.try_into().expect("resolution failed");
-        hir.entries
-            .into_iter()
-            .filter_map(|e| {
-                if let resolution::Entry::Transaction(txn) = e.data {
-                    Some(txn)
-                } else {
-                    None
-                }
-            })
-            .collect()
+        hir.transactions().collect()
     }
 
     #[test]
@@ -201,11 +192,10 @@ mod write_ledger_tests {
         write_ledger([txn], &mut out).unwrap();
         let text = String::from_utf8(out).unwrap();
 
-        assert!(text.starts_with("2024-01-15 Groceries"));
-        assert!(text.contains("Expenses:Food"));
-        assert!(text.contains("Assets:Checking"));
-        // No trailing blank line for a single transaction.
-        assert!(!text.ends_with("\n\n"));
+        assert_eq!(
+            text,
+            "2024-01-15 Groceries\n    Expenses:Food  50 $\n    Assets:Checking\n"
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -82,3 +149,200 @@ where
...
+
+        let mut out: Vec<u8> = Vec::new();
+        write_ledger([txn], &mut out).unwrap();
+        let text = String::from_utf8(out).unwrap();
+
+        assert!(text.starts_with("2024-01-15 Groceries"));
+        assert!(text.contains("Expenses:Food"));
+        assert!(text.contains("Assets:Checking"));
+        // No trailing blank line for a single transaction.
+        assert!(!text.ends_with("\n\n"));
```
alevy: These checks along with similar ones in the other tests might be better as a full expected text equality comparison for the output, rather than checking for individual substring appearance.

src/main.rs
```diff
@@ -352,11 +352,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
             let hir: ledger::resolution::HIR = ast_journal.try_into()?;
-            for entry in hir.entries {
-                if let ledger::resolution::Entry::Transaction(txn) = entry.data {
-                    println!("{txn}");
+            let txns = hir.entries.into_iter().filter_map(|e| {
+                if let ledger::resolution::Entry::Transaction(txn) = e.data {
+                    Some(txn)
+                } else {
+                    None
                 }
```
alevy: This should probably be put into a helper method on the `HIR` type `transactions()` that returns an iterator over only the transaction entries. It's not strictly part of the funcationality in this PR, but seems sufficiently warrented for making this PR an actual ergonomic improvement.

PR #75, opened 2026-04-27
feat: support glob patterns in include directives

### LINKED ISSUE #73: Support glob patterns in !include directives (REQ-GAP-004b)
## Description

Support glob pattern expansion in `include` (and `!include`) directives.

This is **REQ-GAP-004b** from `docs/requirements.md` (line 496) — a confirmed need from the `betterbytes-org/ledger` migration.

## Use Case

`betterbytes-org/ledger` uses `!include ../people/*.ledger` to include all per-person account files. Without glob support, the migration to doppio is blocked or requires explicit `include` lines for every file.

## Current State

The `include` directive supports single file paths only. The CLI's `file_opener` helper in `src/lib.rs` already uses `glob::glob()`, but this isn't wired through the parser's include handling.

## Required Behavior

- Glob pattern expansion using standard `*`, `?`, `[...]` syntax
- **Deterministic ordering**: lexicographic sort of matched paths (consistent across platforms)
- **Error on zero matches**: a glob that matches nothing should fail with a clear error, not silently include nothing
- Relative paths resolved against the directory of the file containing the directive

## Acceptance Criteria

- [ ] `include foo/*.ledger` includes all matching `.ledger` files
- [ ] Files included in lexicographic order (deterministic across runs and platforms)
- [ ] Glob with no matches produces an error pointing to the source location
- [ ] Existing single-file `include` continues to work unchanged
- [ ] Test coverage for: single file, multiple matches, no matches, recursive glob (`**`)

## References

- `docs/requirements.md` §5.1.3, REQ-GAP-004b
- Existing glob implementation in `src/lib.rs:file_opener`
- Phase 4 milestone (umbrella: #17 / #23 — though this is independent of both)

alevy:
## Summary

Closes #73

Implements glob pattern support for `include` directives per REQ-GAP-004b.

- **Glob expansion**: `include accounts/*.ledger` expands all matching files via the `glob` crate already in the dependency tree.
- **Deterministic ordering**: Matched paths are explicitly sorted after expansion, guaranteeing lexicographic order across all filesystems (POSIX makes no ordering guarantee for directory iteration).
- **Error on zero matches**: A glob pattern containing `*`, `?`, or `[` that matches no files returns a clear error rather than silently including nothing. A literal path that does not exist also errors (previously would silently include nothing).
- **Recursive glob**: `include people/**/*.ledger` works via `glob`'s `**` support.

## Breaking API change (pre-1.0)

`Parser<F>` opener signature and `parse_ledger`/`Parser::parse` return types changed:

| Before | After |
|--------|-------|
| `F: Fn(&str) -> String` | `F: Fn(&str) -> Result<String, Box<dyn std::error::Error>>` |
| `Parser::parse` returns `Result<Journal, pest::error::Error<Rule>>` | returns `Result<Journal, Box<dyn std::error::Error>>` |
| `parse_ledger` returns `Result<Journal, pest::error::Error<Rule>>` | returns `Result<Journal, Box<dyn std::error::Error>>` |

Migration: change opener closures from `|_| String::new()` to `|_| Ok(String::new())`.

`?` and `.unwrap()` callsites need no changes.

## Design decisions

- **`Box<dyn Error>` over a custom error enum**: The opener is a user-supplied callback; a custom enum would require every user-defined error type to be wrapped. `Box<dyn Error>` is the minimal change that threads errors through cleanly without constraining callers.
- **Explicit sort after expansion**: The `glob` crate docs say "paths are visited in an unspecified order". Sorting after collection guarantees deterministic, cross-platform lex ordering.
- **Glob logic in `file_opener`, not `Parser`**: Keeps the parser I/O-agnostic. Library users who implement custom openers can apply their own glob semantics (or none).

## Test plan

- [x] `tests/glob_include.rs`: single-file literal include, multi-file glob in lex order, zero-match glob error, literal-path-not-found error, recursive `**` glob
- [x] All existing inline tests, `tests/dop_format.rs`, and all benchmarks updated to use the new `Ok(String::new())` opener form

### DIFF
--- a/benches/parse.rs
+++ b/benches/parse.rs
@@ -8,7 +8,7 @@ fn bench_parse(c: &mut Criterion) {
     group.measurement_time(Duration::from_secs(15));
     for (name, input) in data::workloads() {
         let mut parser = doppio::parser::Parser {
-            opener: |_: &str| String::new(),
+            opener: |_: &str| Ok(String::new()),
             base_path: PathBuf::new(),
         };
         group.bench_with_input(BenchmarkId::new("parse", name), &input, |b, input| {
--- a/benches/pipeline.rs
+++ b/benches/pipeline.rs
@@ -3,8 +3,8 @@ use std::{path::PathBuf, time::Duration};
 
 mod data;
 
-fn make_parser() -> doppio::parser::Parser<impl Fn(&str) -> Result<String, Box<dyn std::error::Error>>>
-{
+fn make_parser()
+-> doppio::parser::Parser<impl Fn(&str) -> Result<String, Box<dyn std::error::Error>>> {
     doppio::parser::Parser {
         opener: |_: &str| Ok(String::new()),
         base_path: PathBuf::new(),
--- a/benches/pipeline.rs
+++ b/benches/pipeline.rs
@@ -3,9 +3,10 @@ use std::{path::PathBuf, time::Duration};
 
 mod data;
 
-fn make_parser() -> doppio::parser::Parser<impl Fn(&str) -> String> {
+fn make_parser() -> doppio::parser::Parser<impl Fn(&str) -> Result<String, Box<dyn std::error::Error>>>
+{
     doppio::parser::Parser {
-        opener: |_: &str| String::new(),
+        opener: |_: &str| Ok(String::new()),
         base_path: PathBuf::new(),
     }
 }
--- a/benches/queries.rs
+++ b/benches/queries.rs
@@ -4,8 +4,8 @@ use std::{collections::BTreeMap, path::PathBuf, time::Duration};
 
 mod data;
 
-fn make_parser() -> doppio::parser::Parser<impl Fn(&str) -> Result<String, Box<dyn std::error::Error>>>
-{
+fn make_parser()
+-> doppio::parser::Parser<impl Fn(&str) -> Result<String, Box<dyn std::error::Error>>> {
     doppio::parser::Parser {
         opener: |_: &str| Ok(String::new()),
         base_path: PathBuf::new(),
--- a/benches/queries.rs
+++ b/benches/queries.rs
@@ -4,9 +4,10 @@ use std::{collections::BTreeMap, path::PathBuf, time::Duration};
 
 mod data;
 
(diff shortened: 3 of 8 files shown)


alevy:
Thanks for the review.

**Addressed:**
- Docstring inaccuracy about "absolute path" — fixed in fbef32d.

**Acknowledged but not changing in this PR:**
- **`to_str().unwrap()` panic on non-UTF-8 paths** — this code already existed pre-PR; the change here only added `?` for opener errors, not the path conversion. Will leave as-is to keep this PR scoped to glob expansion. Worth a follow-up issue if non-UTF-8 ledger paths become a real consumer concern.
- **Suppressed comment: `new_base_path` double-join** — same situation: the `include_path.parent().map(|p| self.base_path.join(p))` logic was unchanged by this PR. The double-join concern is pre-existing. Tested cases (single literal include, recursive globs across subdirs) work correctly in the new integration tests, so the practical behavior is fine. Worth filing separately if a real failure mode emerges.
- **Suppressed comment: glob includes set base_path once for files in different subdirs** — valid concern for nested includes inside glob-matched files, but no consumer hits this today (`betterbytes-org/ledger`'s glob usage is flat: `!include ../people/*.ledger`). Would justify the larger refactor of having the opener return per-file paths-and-contents only when a real use case emerges.

alevy:
Pushed in 064992d:

**Fixed (real bug):**
- File concatenation lacked a separator. Now inserts `\n` between files when the buffer doesn't already end with one. New test `glob_files_without_trailing_newline_are_separated` covers this.

**The other three new comments are wrong (CI confirms):**
- *"`format!(...).into()` won't compile"* — `Box<dyn Error>: From<String>` is in stdlib (the `String: From<String>` impl + the `Box<dyn Error>: From<E: Error>` impl, plus a specific `Box<dyn Error>: From<String>` blanket). The code compiles and CI runs the tests.
- *"`pattern.contains(['*', '?', '['])` won't compile"* — `str::contains` accepts `&[char]` / `[char; N]` patterns since Rust 1.51. Confirmed working on the pinned 1.95 toolchain.
- *"non-UTF-8 path panic"* — already responded to: this code is unchanged from pre-PR; out of scope here.

PR #77, opened 2026-04-27
fix: infer commodity for bare balance assignments from account running balance (#71)

### LINKED ISSUE #71: Balance assignment without commodity (=0) fails when account has no commodity context
## Description

Postings with equality/balance assignment syntax without an explicit amount (e.g., `=0`) fail with `AmountWithNoCommodity` when there's no commodity specified or inferred in context.

## Minimal Failing Case

```ledger
2026-04-01 Test
    Account A  $100
    Account B  =0
```

Running `ledger balance` on this produces:
```
Error: AmountWithNoCommodity
```

## Expected Behavior

According to Ledger documentation and OG ledger-cli behavior, `=0` is a balance assignment: it should infer the amount needed to bring Account B to balance 0. The commodity should be inferred from:

1. Account B's existing running balance (if it has prior postings)
2. Other postings in the same transaction (if they establish a commodity context)
3. A default commodity directive (if set)

## Current Behavior

The elaborator attempts to evaluate `0` as a value expression but has no commodity context. It looks for a default commodity directive and fails when none exists, producing `AmountWithNoCommodity`.

## Root Cause

In `elaboration.rs:456`, `eval_and_normalize_amount` is called on the assignment expression. The evaluator reduces `0` to `ValueExpr::Amount { value: 0, commodity: None }`, and on line 629, when there's no commodity and no default, it errors.

For balance assignments specifically, the commodity should be inferred from the account's current balance context before evaluating the target expression.

## Test Cases to Verify

- [ ] `Account =0` when account has no prior balance and no commodity context → should error with clearer message
- [ ] `Account =0` when account has prior balance of `00` → should infer commodity from account balance  
- [ ] `Account =0` with default commodity set → should use default commodity
- [ ] Multi-commodity account with `=0` → needs clarification on expected behavior

alevy:
## Summary

- Bare `=TARGET` balance assignment postings (e.g. `=0`) now correctly infer the commodity from the account's existing running balance when no commodity is given and no default commodity directive is active
- When the account holds a balance in exactly one commodity, that commodity is used as the fallback — the common single-commodity case
- The fix adds a new internal `eval_and_normalize_amount_with_fallback` function that accepts an optional caller-supplied fallback commodity; `eval_and_normalize_amount` delegates to it with `None` (no change to existing callers)
- Multi-commodity accounts with a bare assignment and no default commodity still error as before — the ambiguous case cannot be resolved automatically

## Test plan

- [x] `=0` after a prior `Account $100` posting succeeds and produces a -$100 delta
- [x] `=$0` with explicit commodity still works as before
- [x] `=0` with a `default commodity` directive still works (no regression on existing path)
- [x] `=0` on an account with no prior balance and no default commodity still errors with `AmountWithNoCommodity`
- [x] `cargo test` — all 76 tests pass
- [x] `cargo clippy` — no warnings
- [x] `cargo fmt` — applied

Closes #71

### DIFF
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -453,11 +453,31 @@ impl TryFrom<resolution::HIR> for Journal {
                                 AmountDetails::BalanceAssignment(assignment) => {
                                     // "= target_balance" — compute the delta needed to reach the
                                     // target from the current running balance.
-                                    let (newsum, commodity) = evaluator::eval_and_normalize_amount(
-                                        assignment,
-                                        entry_context,
-                                        &state,
-                                    )?;
+                                    //
+                                    // If the assignment expression is bare (no commodity), try to
+                                    // infer the commodity from the account's existing running
+                                    // balance. This is the common case: `= 0` after the account
+                                    // already holds a balance in a single commodity.
+                                    let inferred_commodity = account_balance.and_then(|ab| {
+                                        let mut keys = ab.commodity.keys();
+                                        let first = keys.next()?;
+                                        // Only infer if there is exactly one commodity — if
+                                        // the account holds multiple commodities a bare `=0`
+                                        // is ambiguous and we fall through to the normal
+                                        // error path.
+                                        if keys.next().is_none() {
+                                            Some(first.as_str())
+                                        } else {
+                                            None
+                                        }
+                                    });
+                                    let (newsum, commodity) =
+                                        evaluator::eval_and_normalize_amount_with_fallback(
+                                            assignment,
+                                            entry_context,
+                                            &state,
+                                            inferred_commodity,
+                                        )?;
                                     let value = newsum
                                         - account_balance
                                             .and_then(|ab| ab.commodity.get(&commodity))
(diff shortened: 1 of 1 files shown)


alevy:
Is this the correct behavior? As opposed to expanding the assertion to be on all possible commodities or something?

alevy:
Good question. I think the single-commodity behavior is correct, for a few reasons:

1. **OG ledger semantics**: `=0` in ledger-cli is shorthand for `=$0` (or whatever single commodity is in scope) — it doesn't auto-expand to all commodities the account holds. So matching this preserves compatibility.

2. **AST invariant**: Each posting has exactly one `AmountDetails`. Expanding a bare `=0` into multiple deltas across commodities would mean one source line generates multiple AST postings, which breaks that invariant and complicates everything downstream (transaction balance, null-posting inference, the elaborated journal shape).

3. **Genuine ambiguity**: If an account holds USD and EUR, what does `=0` mean? Zero out both at the current FX rate? Zero out each independently (creating two postings)? There's no single "obvious" answer — better to error and force the user to be explicit.

If a user wants to zero out a multi-commodity account, they can write the postings explicitly:
```
2026-04-04 Drain
    Account A  =$0
    Account A  =EUR 0
    ...
```

I added a regression test for the zero-balance-pruning case Copilot flagged (a stale $0 entry shouldn't make a single-EUR-balance account look ambiguous). Pushed in 47bb9a4.

alevy:
Good catch — the original issue #71 explicitly listed three fallback levels (account balance, same-transaction context, default commodity). The first PR only implemented (1) and (3); the same-transaction fallback was deferred but is small enough to land here too.

Implemented in 4464914:
- After the account-balance check fails to infer, look at `transaction_state` (the per-commodity totals of postings already processed in this transaction). If exactly one commodity, use it.
- New test `test_balance_assignment_infers_commodity_from_same_transaction` covers this — verifies the original issue's minimal failing case (`Account A $100 / Account B =0 / Account C $-100`) now elaborates without `AmountWithNoCommodity`.
- Old `test_balance_assignment_no_context_errors` simplified to a single bare-`=0` posting (no other context), since the prior shape is now valid behavior.

Now actually closes #71.

PR #86, opened 2026-04-27
feat: regex match operators (=~ / !~) and tag() function

### LINKED ISSUE #79: Regex match operators (=~, !~) in value expressions
## Description

Support `=~` (regex match) and `!~` (regex not-match) as comparison operators in `value_expr` / `bool_expr`.

Tag validation and account assertions require regex matching:
```
tag Statement
    assert value =~ /[^\/].+/

define incomeChecker(amt) = tag("Entity") !~ /^\s*$/ and ...
```

## Required

- New cmp_op variants: `=~`, `!~`
- Regex literal syntax: `/pattern/` (consider escaping rules)
- Evaluator: match string LHS against regex RHS; `Str =~ Regex → bool`
- Compile pattern once and cache (bb-ledger has many tag rules; per-posting eval would compile repeatedly)

## Foundation

This is a prerequisite for tag directive validation and parameterized define macros (which use `tag("X") !~ /pattern/`).

### LINKED ISSUE #80: tag() function for accessing posting/transaction tag values in expressions
## Description

Support `tag("name")` as a function call in value expressions, returning the tag's value as a string (or empty string if absent).

```
define incomeChecker(amt) = tag("Entity") !~ /^\s*$/ and tag("IncomeType") !~ /^\s*$/

account Income:Grants
    assert incomeChecker(amount)
```

## Required

- Grammar: `tag("name")` parses as a function call (already supported via `function_call` rule, just needs binding)
- Evaluator: register `tag` as a built-in callable in the elaboration context
- Lookup: tag value comes from the *current posting* being evaluated (or the transaction). Returns "" if not present.

## Depends on

#79 (regex match) — typical tag() usage is paired with `=~`/`!~`.

alevy:
## Summary

Implements two expression primitives used together in `bb-ledger` account directives:

- **`=~` and `!~` (issue #79)** — regex comparison operators in `bool_expr` context. The RHS must be a `/pattern/` regex literal. Escaped slashes (`\/`) inside the pattern are supported. Regex matching is case-sensitive by default (Rust `regex` crate semantics, consistent with how the regex crate behaves out of the box).

- **`tag("name")` (issue #80)** — a new built-in function that looks up a posting's metadata key (parsed from `; Key: Value` notes by the resolution stage) and returns its string value. If the key is absent, an empty string is returned — so `tag("X") !~ /^\s*$/` works correctly for blank-tag detection without requiring a separate null check.

### Representative real-world usage

```ledger
account Income:Salary
    assert tag("Entity") !~ /^\s*$/ and tag("IncomeType") !~ /^\s*$/

account Assets:Checking
    assert value =~ /[^\/].+/
```

## Design decisions

- **Regex RHS is grammar-restricted to `regex_literal`** — following OG ledger convention and keeping the grammar unambiguous. The `/` character cannot appear as the start of an ordinary expression, so adding a separate `regex_cmp_op` branch in `bool_expr` avoids any grammar conflict with the division operator.
- **Pattern compiled per call** — the `regex` crate is fast enough for the assertion workload (few patterns, called per-posting). A compilation cache can be added later if profiling shows it as a hot path.
- **`eval` gains a `posting_metadata` parameter** — threaded through all recursive calls so `tag()` has access without a global side-channel. `eval_and_normalize_amount` passes an empty `BTreeMap` since amount expressions never involve metadata.
- **`ValueExpr::Regex` is NOT compiled at parse time** — the AST stays `regex`-crate-free. The evaluator compiles on use.

## Test plan

- [x] `assert "abc" =~ /^a/` passes, `assert "abc" =~ /^z/` fails
- [x] `assert "abc" !~ /^z/` passes, `assert "abc" !~ /^a/` fails
- [x] Commodity string from `commodity` variable matched against regex
- [x] `tag("X") =~ /^foo/` passes when posting has `; X: foobar`
- [x] `tag("X") !~ /^foo/` passes when posting has no X tag (empty string)
- [x] `tag("X") =~ /^foo/` fails when posting has no X tag
- [x] Chained `tag("A") !~ /^\s*$/ and tag("B") !~ /^\s*$/` — both present passes, one missing fails
- [x] All 89 pre-existing tests continue to pass (99 total after adding 10 new ones)
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt` clean

Closes #79
Closes #80

### DIFF
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -25,7 +25,7 @@
 //!   to [`Journal::accounts`], merging any properties declared in `account`
 //!   directives.
 
-use std::{cell::RefCell, collections::BTreeMap, collections::HashMap, fmt::Display};
+use std::{collections::BTreeMap, fmt::Display};
 
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
@@ -89,18 +89,9 @@ struct AccountBalances {
 /// Account balances are updated as each transaction is processed so that
 /// balance assertions and the `account()` function see the balance *before*
 /// the current posting is applied (which matches ledger-cli semantics).
-///
-/// `regex_cache` memoises compiled regexes across all postings within one
-/// elaboration run. A typical bb-ledger workload has ~10 distinct patterns
-/// repeated over hundreds of postings; compiling each once rather than per
-/// evaluation reduces compilation overhead to O(distinct patterns).
 #[derive(Default, Clone, Debug)]
 struct RunningState {
     account_balances: BTreeMap<String, AccountBalances>,
-    /// Cache of compiled regexes keyed by pattern string.
-    /// `RefCell` provides interior mutability so the cache can be populated
-    /// from `eval_cmp`, which receives only a shared reference to `RunningState`.
-    regex_cache: RefCell<HashMap<String, regex::Regex>>,
 }
 
 /// A fully evaluated and balanced transaction, ready for serialisation.
@@ -873,7 +864,7 @@ mod evaluator {
                     ast::ValueExpr::Regex(_) => rhs_expr.clone(),
                     other => eval(other.clone(), &ctx, state, posting_metadata)?,
                 };
-                eval_cmp(cmp_op, &lhs_val, &rhs_val, state)?
+                eval_cmp(cmp_op, &lhs_val, &rhs_val)?
             }
         };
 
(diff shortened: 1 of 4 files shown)


alevy:
> ValueExpr::Regex is NOT compiled at parse time — the AST stays regex-crate-free. The evaluator compiles on use.

This suggests that it's possible for the parser to succeed even if the regular expression is syntactically incorrect, doesn't it? Keeping the AST regex-crate free doesn't seem like a particularly important goal.

src/elaboration.rs
```diff
@@ -89,9 +89,18 @@ struct AccountBalances {
...
+/// elaboration run. A typical bb-ledger workload has ~10 distinct patterns
+/// repeated over hundreds of postings; compiling each once rather than per
+/// evaluation reduces compilation overhead to O(distinct patterns).
 #[derive(Default, Clone, Debug)]
 struct RunningState {
     account_balances: BTreeMap<String, AccountBalances>,
+    /// Cache of compiled regexes keyed by pattern string.
+    /// `RefCell` provides interior mutability so the cache can be populated
+    /// from `eval_cmp`, which receives only a shared reference to `RunningState`.
+    regex_cache: RefCell<HashMap<String, regex::Regex>>,
```
alevy: Adding a `RefCell` is a real commitment and brings with it re-entrancy problems and complexity. Using it should really be justified by a measurable performance improvement, and shouldn't be done just in anticipation.

alevy:
Addressed in 99e6f97:

**Both concerns fixed without compromise:**

1. **RefCell cache removed.** Eval re-compiles per call now — back to the original (pre-cache) behavior. No interior mutability, no re-entrancy concerns. Agree this shouldn't be added speculatively; if bb-ledger or another workload shows it's a hotspot we can revisit with measurements.

2. **Regex validated at parse time.** Added `validate_regexes` walker in `parser.rs` that runs after the AST is built. Invalid patterns like `/[unclosed/` now fail `parse_ledger` with a message identifying the pattern, rather than waiting until elaboration evaluates the assertion.

Considered storing the compiled regex in the AST (per your point 2 hint), but `regex::Regex` doesn't impl `Eq` so we'd need a custom PartialEq wrapper that propagates through every `ValueExpr` consumer (parser tests, Display impls, etc.). Validation-without-storing keeps the AST simple — the small cost is that a valid pattern is compiled twice (once for validation, once per eval). If that becomes meaningful we can store the compiled form behind a newtype.

Updated the invalid-regex tests to assert on parse-time failure. `InvalidRegexPattern` evaluation variant kept as a defensive fallback (the test that exercises it is gone, but if future code paths bypass validation it still produces a clear error).

Issue #81, opened 2026-04-27
Parameterized define macros (REQ-GAP-004d)


alevy:
## Description

Extend `define` directive to support parameter lists, enabling reusable predicates and helpers.

## Current state

```
define name = expr        ; works today (single expression substitution)
define name(param) = expr ; NOT yet supported
```

## Use case (betterbytes-org/ledger)

```
define assetChecker(amt) = (amt > -100.00 or (tag("TaxImplication") !~ /^\s*$/ and tag("Entity") !~ /^\s*$/))
define expenseChecker(amt) = (amt < $100 or tag("Receipt") !~ /^\s*$/ or ...)
define incomeChecker(amt) = ...

account Income:Grants
    assert incomeChecker(amount)
```

## Required

- Grammar: `define name(param1, param2, ...) = expr` — parameter list in parens
- AST: `Define` directive carries optional parameter list
- Resolution: store as callable in `Context::defines`
- Evaluator: when an identifier is invoked as `f(args)`, look up `defines[f]`, substitute parameters, evaluate

## Depends on

#79 (regex match), #80 (`tag()` function) — almost all real defines use both.

## Part of

REQ-GAP-004d (split issue: define macros)

PR #87, opened 2026-04-27
feat: tag directive value validation

### LINKED ISSUE #82: Tag directive value validation (REQ-GAP-004d)
## Description

Properly parse `assert` / `check` subdirectives nested under `tag` directives, where `value` is bound to the tag's string value.

## Current state

The grammar's `tag_directive` rule has a permissive body that accepts but ignores any indented content. No validation happens.

```
tag Statement
    assert value =~ /[^\/].+/
    check value =~ /[^\/][^ ]+(\/[^ ])+/

tag IncomeType
    assert value =~ /^(Donations|RBI|UBTI)$/
```

These declare tag schemas: every transaction or posting carrying `Statement: <value>` must have `<value>` match the regex.

## Required

- Grammar: extend `tag_directive` to parse `assert`/`check` lines (analogous to `account_directive`)
- AST: tag directive carries `asserts: Vec<BoolExpr>` and `checks: Vec<BoolExpr>`
- Resolution: store on a per-tag-name properties map
- Elaboration: for every posting/transaction carrying tag `X: V`, evaluate the asserts/checks for tag `X` with `value` bound to `V`. Failed assert halts; failed check warns.

## Depends on

#79 (regex match) — nearly every real tag validation uses `=~`.

## Part of

REQ-GAP-004d (split issue: tag directives)

alevy:
## Summary

- Extends the `tag` directive grammar to accept `assert` and `check` sub-directives with full `bool_expr` bodies (mirrors `account_directive`)
- Parses `tag_assert` / `tag_check` into a new `Directive::Tag { asserts, checks }` AST variant
- Accumulates tag validation rules in `GlobalContext::tag_properties` (a new `BTreeMap<String, TagProperties>`) during resolution
- During elaboration, validates metadata-style tags (`; TagName: value`) on both transaction-level and posting-level metadata by binding `value` as a `Str` in the expression context; failed `assert` returns `ElaborationError::TagAssertionFailed`; failed `check` warns to stderr
- Bare colon-tags (`; :payroll:`) are intentionally skipped — they carry no value
- Adds 9 targeted tests covering assert pass/fail, check warn-and-continue, posting-level validation, bare-tag bypass, and the declared-but-unused case
- Also fixes the in-progress parameterized-define work (already on this branch) to compile cleanly

Closes #82

## Test plan

- [x] `tag X\n    assert value =~ /^foo/` with `; X: foobar` → passes
- [x] Same with `; X: barfoo` → `TagAssertionFailed`
- [x] `tag X\n    check value =~ /^foo/` with `; X: barfoo` → warns, elaboration succeeds
- [x] Multiple asserts under one tag (IncomeType regex) → all must pass
- [x] Tag declared but not referenced → no errors
- [x] Posting-level metadata validated (not just transaction-level)
- [x] Bare colon-tags (`; :payroll:`) NOT validated — no value to check
- [x] All 113 unit tests + integration tests pass

### DIFF
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -318,6 +318,9 @@ pub enum EvaluationError {
         /// Number of arguments provided at the call site.
         got: usize,
     },
+    /// Expression evaluation exceeded the recursion limit. The most common
+    /// cause is mutually-recursive defines, e.g. `define a = b; define b = a`.
+    RecursionLimitExceeded,
 }
 
 impl Display for EvaluationError {
@@ -370,6 +373,12 @@ impl Display for EvaluationError {
                 write!(
                     f,
                     "define '{name}' expects {expected} argument(s), got {got}"
+                )
+            }
+            EvaluationError::RecursionLimitExceeded => {
+                write!(
+                    f,
+                    "expression evaluation exceeded recursion limit (likely a cyclic `define`)"
                 )
             }
         }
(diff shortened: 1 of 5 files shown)


alevy:
Addressed the stack-overflow blocker in b3f636c.

```
define a = b
define b = a
```
now produces `EvaluationError::RecursionLimitExceeded` (with a clear Display message identifying the likely cause) rather than crashing the process. Implemented as a thread-local depth counter with an RAII guard at `eval` entry, capped at 64 (well above real-world expression depth, well below debug-build stack limits). Tests cover both self-referential (`define x = x`) and mutually-recursive cases.

**Other review concerns acknowledged but deferred:**

- *`define_body = bool_expr | value_expr` value branch is dead* — grammar nit, fine as-is for now (the parser handles both shapes uniformly via downstream detection); could be cleaned up in a follow-up if it bothers anyone.
- *Zero-param bool-body define used as bare identifier silently does nothing* — the case `define inBounds = amount > 0; assert inBounds` is genuinely surprising, but rare in practice (bb-ledger always uses `assert inBounds(amount)` form). Worth a follow-up issue rather than blocking this PR.
- *`eval_bool_expr` / `eval_bool_expr_with_context` chain-handling duplication* — pure refactor, follow-up.

Note: this PR's branch contains commits closing both #81 and #82. PR #88 (which stacks on this branch with a small refactor only) is now redundant; closing it.

src/elaboration.rs
```diff
@@ -1409,12 +1418,45 @@ mod evaluator {
...
+    /// Maximum recursion depth for [`eval`]. Protects against cyclic
+    /// `define`s (e.g. `define a = b; define b = a`), which would otherwise
+    /// recurse until the OS aborts the process with a stack overflow.
+    /// 64 is well above any sane real-world expression depth and well below
+    /// debug-build stack limits (debug frames are large).
+    const MAX_EVAL_DEPTH: usize = 64;
+
+    thread_local! {
+        static EVAL_DEPTH: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+    }
```
alevy: Whoa! `thread_local!` is a huge red flag! Why is there a global here? Why not just pass in an eval depth to recursive calls that errors when it reaches 0?

alevy:
Fair call — addressed in cb0c89f. Replaced the thread_local + RAII guard with an explicit `budget: usize` parameter on `eval()`. Each external caller passes `EVAL_BUDGET` (= 64); each recursive `eval` call passes `budget` unchanged after `eval`'s entry decrements it via `budget.checked_sub(1)`. Same protection, but the depth is now visible in the function signature.

PR #88, opened 2026-04-27
feat: parameterized define macros

### LINKED ISSUE #81: Parameterized define macros (REQ-GAP-004d)

alevy:
## Summary

- Extends `define` directives to accept an optional parameter list: `define myfilter(x) = x > 100`
- Supports both value-returning and bool-returning define bodies
- Bool-body defines can be called in filter expressions like functions: `myfilter(amount)`
- Arguments are evaluated in the caller context, then bound by name in a child context where the define body is evaluated
- Backward-compatible: zero-param defines continue to work as before

Closes #81

## Test plan

- [ ] `cargo test` passes (8 new tests for parameterized defines)
- [ ] `cargo clippy` clean
- [ ] `cargo fmt` clean
- [ ] Zero-param defines (existing behavior) still pass all existing tests
- [ ] Multi-arg define with both value and bool bodies evaluated correctly
- [ ] Bool-body define callable in filter position
- [ ] Arg count mismatch returns `DefineArgCountMismatch` error
- [ ] Param names shadow outer defines correctly

### DIFF
--- a/src/elaboration.rs
+++ b/src/elaboration.rs
@@ -1005,72 +1005,70 @@ mod evaluator {
         // Check whether the LHS is a call to a bool-body define. If it is and
         // there is no comparison operator, we expand the define body as a full
         // bool expression rather than treating the call result as a numeric value.
-        if expr.cmp.is_none() {
-            if let ast::ValueExpr::Function { ref name, ref args } = expr.lhs {
-                if let Some(define) = ctx.defines.get(name.as_str()) {
-                    if let ast::DefineBody::Bool(ref body) = define.body.clone() {
-                        if define.params.len() != args.len() {
-                            return Err(EvaluationError::DefineArgCountMismatch {
-                                name: name.clone(),
-                                expected: define.params.len(),
-                                got: args.len(),
-                            });
-                        }
-                        // Bind arguments into a new context and evaluate the bool body.
-                        let mut call_ctx = ctx.clone();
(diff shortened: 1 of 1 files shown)


alevy:
Closing as redundant — the parameterized define commit (closes #81) is included in PR #87's branch alongside the tag directive validation commit (closes #82). Merging #87 will close both issues. This stacked PR's only diff was a small refactor to the same code, which I'll port to #87 if it's wanted.

Issue #101, opened 2026-04-27
[Phase A] Add wasm32-unknown-unknown to CI build matrix


alevy:
## Goal

Add a CI job that builds the doppio library for `wasm32-unknown-unknown` so any future regression (a new dep that doesn't compile to WASM) is caught at PR time.

**Blocked on**: #99 (audit) and #100 (xz removal). Once both are in, this is a CI-only change.

## What to change

`.github/workflows/ci.yml`: add a `wasm-build` job that:

```yaml
wasm-build:
  name: WASM library build
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v4
    - uses: dtolnay/rust-toolchain@master
      with:
        toolchain: "1.95.0"
        targets: wasm32-unknown-unknown
    - uses: Swatinem/rust-cache@v2
    - run: cargo build --target wasm32-unknown-unknown --lib
```

If the audit (#99) identifies dep features that need to be turned off for WASM, parameterise the build with the appropriate `--no-default-features` / feature flags.

## Out of scope

- Building examples or binaries for WASM (the binary uses `clap`, `fs::File`, etc.; library-only is sufficient).
- Publishing wasm-pack-style packages.
- Browser-targeted wrappers (post-1.0 / doppio-web).

## Acceptance

- [ ] CI green on a known-good commit
- [ ] CI red on a contrived dep that doesn't compile to WASM (verify by temporarily breaking, then revert)

Issue #100, opened 2026-04-27
[Phase A] Migrate Journal serialization from postcard+xz to canonical protobuf


alevy:
## Goal

Swap the `.dop` body from postcard+xz to canonical protobuf as defined in `proto/doppio.proto`. Drop the `postcard`, `xz`, and related deps from the runtime tree.

**Blocked on**: #98 (schema must be locked first).

This is the meatiest issue of Phase A — the actual wire-format migration. Should be its own PR with focused review.

## What to change

- `Cargo.toml`: add `prost` (runtime), `prost-build` (build-time). Drop `postcard` and `xz`.
- `build.rs`: invoke `prost-build` against `proto/doppio.proto`; emit Rust types into `OUT_DIR`.
- `src/lib.rs`: `dop_write_header` / `dop_read_header` bump to format version 3. Replace the postcard `from_io` / `to_io` calls with `prost::Message::encode` / `Message::decode`. Drop xz wrapping.
- `src/elaboration.rs`: replace the existing `Journal` / `ResolvedTransaction` / `ResolvedPosting` / `AccountProperties` / `CommodityProperties` / `Amount` etc. structs with newtypes around the prost-generated types **OR** keep the existing structs and add `From` impls in both directions. Pick one and document the choice in the PR.
- `src/main.rs`: `Compile`/`Balance`/`Register`/`Print`/`Stats`/`Accounts`/`Commodities` consumers all read the new format.
- `tests/fixtures/v1.dop`: rename to `v0.1.0.dop`. Add `tests/fixtures/v0.2.0.dop` (postcard+xz format). Both should produce a clean version-mismatch error against the v0.3 binary.
- Update or replace the `dop_format.rs` integration tests to round-trip via the new format.
- Update `CHANGELOG.md`: add to Unreleased section noting the breaking change.
- `README.md` / `docs/SUPPORTED_FEATURES.md`: update any mention of postcard or xz.

## Definition of done

- [ ] All existing tests pass (`cargo test`)
- [ ] Old `.dop` fixtures rejected with a clear version-mismatch error
- [ ] bb-ledger end-to-end smoke test still passes (compile + balance from `.dop` and from source)
- [ ] No regressions in clippy or doc warnings
- [ ] CHANGELOG entry documents the breaking change

## Out of scope

- Schema design (was #98).
- WASM build target (separate issue: #101).
- Cross-language SDK / docs for consumers (post-1.0).

PR #106, opened 2026-04-27
docs: WASM compatibility audit (closes #99)

### LINKED ISSUE #99: [Phase A / lock-in] Audit dependencies for wasm32-unknown-unknown compatibility
## Goal

Catalog every dep in `Cargo.toml` and verify it builds for `wasm32-unknown-unknown`. Identify any that don't (`xz` is the known one) and propose an alternative or a feature-flagged exclusion.

This is the **second early-lock-in item** for Phase A. Output is a markdown report committed to the repo (e.g. `docs/WASM_AUDIT.md` or as a comment on this issue) — not code changes.

Can run **in parallel with #98** (the schema lock-in). Doesn't block #98 or vice versa.

## What to check

For each direct dependency in `Cargo.toml`:

- [ ] `chrono` (default-features = false; should be fine)
- [ ] `clap` (CLI-only, behind a feature flag for the binary)
- [ ] `glob` (file-system; behind `std::fs` — may need feature gating)
- [ ] `pest` (pure parser, should be fine)
- [ ] `pest_derive` (proc-macro, build-time only)
- [ ] `postcard` (use-std feature — pure Rust; will be removed in #100 anyway)
- [ ] `regex` (pure Rust)
- [ ] `rust_decimal` (pure Rust)
- [ ] `serde` (pure Rust)
- [ ] `serde-pickle` (pure Rust; check whether actually used)
- [ ] `serde_json` (pure Rust)
- [ ] `xz` / `xz2` / `lzma-sys` (**known WASM blocker**; will be removed in #100)
- [ ] `prost` and `prost-build` (pure Rust on the runtime side; build-time only on the build side)

## Method

For each dep:

```sh
nix shell nixpkgs#rustup -c rustup target add wasm32-unknown-unknown
nix shell nixpkgs#rustup -c cargo build --target wasm32-unknown-unknown --no-default-features  # if applicable
```

Or build the whole library after #100 lands (xz removed) and see what fails:

```sh
nix shell nixpkgs#rustup -c cargo build --target wasm32-unknown-unknown --lib
```

## Out of scope

- Adding wasm32 to CI (separate issue: #101).
- Building example WASM consumers (post-1.0).
- Browser-targeted wrappers (separate effort, doppio-web in roadmap).

## Acceptance

- [ ] One bullet per dep listing its WASM status (works, blocker, behind feature flag)
- [ ] For any blocker found, a recommended remediation
- [ ] Report committed to the repo or recorded as the closing comment of this issue

alevy:
## Summary

Resolves #99 (Phase A lock-in): catalogues every direct dependency in `Cargo.toml` against `wasm32-unknown-unknown` and documents the method for re-running the audit.

## Headline

**`xz` is the sole hard blocker.** With only that line commented out of `Cargo.toml`, the entire library compiles cleanly to wasm:

```text
$ cargo build --target wasm32-unknown-unknown --lib
   Compiling doppio v0.2.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.74s
```

`xz` is already slated for removal in #100, so the WASM CI gate (#101) is unblocked once that lands.

## Hygiene findings (non-blocking, suggest follow-up issues)

The audit surfaced three cleanup opportunities that aren't applied here, per #99's 'no code changes' constraint:

- **`serde-pickle` is unused.** Zero references in `src/` or `tests/`. Drop it from `[dependencies]`.
- **`clap` and `serde_json` are binary-only.** Move behind a `cli` feature flag so the library has a smaller surface.
- **`glob` + `file_opener`.** Compile to wasm but can't run usefully in a browser sandbox; consumers should supply their own opener. Optional `#[cfg(not(target_family = "wasm"))]` gate.

## Test plan

Documentation-only PR; CI confirms clippy / tests are unaffected. The audit's method section was exercised on this commit (`4e3baec`).

Closes #99.

### DIFF
--- a/docs/WASM_AUDIT.md
+++ b/docs/WASM_AUDIT.md
@@ -1,0 +1,139 @@
+# WASM compatibility audit
+
+Last verified: 2026-04-26 against rust 1.95.0, wasm32-unknown-unknown.
+
+This audit catalogs every dependency declared in `Cargo.toml` and reports
+its status when targeting `wasm32-unknown-unknown` from the **library**
+crate (`cargo build --target wasm32-unknown-unknown --lib`). The `dop`
+binary is *not* wasm-clean (it uses `clap`, `std::fs`, `xz`, `serde_json`
+for output, etc.) and is not expected to be — only the library surface
+needs to be wasm-buildable so downstream consumers (e.g. `doppio-web`)
+can embed the parser/elaborator in a browser.
+
+Tracking issue: <https://github.com/alevy/doppio/issues/99>.
+
+## Headline result
+
+With **only** `xz` removed from `[dependencies]`, the entire library
+compiles cleanly to `wasm32-unknown-unknown` with no other source or
+feature changes:
+
+```text
+$ cargo build --target wasm32-unknown-unknown --lib
+   Compiling doppio v0.2.0
+    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.74s
+```
+
+All other current direct dependencies are pure-Rust and wasm-clean as
+configured. `xz` is the sole hard blocker, and is already slated for
+removal in #100.
+
+## Library deps
+
+| Crate | WASM status | Notes |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


alevy:
Closing — exploration docs don't belong in the doppio repo itself. The audit is now in the research repo at https://github.com/alevy/doppio-research/blob/main/wasm-audit.md (private). The hygiene findings it surfaced are addressed elsewhere: `serde-pickle` drop and `clap`/`serde_json` binary-only gating are handled by the workspace restructure in #107; `xz` removal is the body of #100.

Issue #117, opened 2026-04-28
[Phase C] Add inherent `proto::Decimal::to_decimal()` method


alevy:
## Goal

`decimal_from_proto(&p)` works but `p.to_decimal()` reads more naturally at call sites and lines up with the convention many Rust users expect. Add:

```rust
impl proto::Decimal {
    pub fn to_decimal(&self) -> rust_decimal::Decimal {
        decimal_from_proto(self)
    }
}
```

Keep `decimal_from_proto` as the free function for code that prefers it; the inherent method is a small ergonomic alias.

## Motivation

NOTES F5 / Concrete suggestions item 3 from `doppio-research/prototypes/doppio-categorize/NOTES.md`. Adjacent to issue #116 (`Display` on `proto::Decimal`); the two together close the proto-decimal ergonomic gap.

## Acceptance

- [ ] `proto::Decimal::to_decimal(&self) -> rust_decimal::Decimal` implemented.
- [ ] `decimal_from_proto` remains public (it's also documented in the proto file as the canonical recipe).
- [ ] Test asserts `p.to_decimal() == decimal_from_proto(&p)` for a few values.
- [ ] Doc comment cross-links the free function and notes both are equivalent.

PR #128, opened 2026-04-28
feat(proto): Amount/Posting accessor methods (closes #114)

### LINKED ISSUE #114: [Phase C] Add proto::Amount and proto::Posting accessor methods (iter, get, amounts, amount_in)
## Goal

Add inherent methods to `proto::Amount` and `proto::Posting` that absorb the F1–F3 friction (Option<Amount> wrapper, HashMap of proto::Decimal, decimal_from_proto reassembly) into a clean accessor API.

```rust
impl proto::Amount {
    /// Iterate (commodity, decimal) pairs. Order is unspecified — the
    /// underlying map is unordered per the protobuf spec.
    pub fn iter(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_ { ... }

    /// Get the decimal value for a specific commodity, if present.
    pub fn get(&self, commodity: &str) -> Option<rust_decimal::Decimal> { ... }
}

impl proto::Posting {
    /// Iterate (commodity, decimal) pairs across this posting's amount.
    /// Empty if the posting has no amount field set.
    pub fn amounts(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_ { ... }

    /// Get the decimal value for a specific commodity in this posting's
    /// amount, if both are present.
    pub fn amount_in(&self, commodity: &str) -> Option<rust_decimal::Decimal> { ... }
}
```

The `Posting` methods compose the `Amount` methods through the `Option<Amount>` wrapper. Implementations are a few lines each, but the surface is what consumer code actually wants to write.

### Why these specific shapes

Real consumers fall into three cases when they need posting amounts:

- **(b) Want all commodities** — printing a detailed balance, formatting a register row that should show every commodity in a multi-commodity posting. Use `amounts()` / `iter()`.
- **(c) Want a specific commodity** — bb-ledger / bookie importing USD-typed Mercury transactions, doppio-categorize matching against USD queries, CLI users filtering with `--currency USD`. Use `amount_in(commodity)` / `get(commodity)`.
- **(a) Want all values converted to a target commodity** — dashboards, runway calculations, "total in USD." This requires `journal.prices` lookup, date-relative price selection, possibly FX cross-rates. It's a genuinely higher-level operation, not a per-Posting accessor — see "Out of scope" below.

A previous version of this issue proposed a `single_amount() -> Option<...>` returning "the only commodity if exactly one is present." Discussion in the comments showed that's a degenerate primitive: it serves no real consumer use case (none of (a)/(b)/(c) want "any one commodity if only one"), and it relies on undefined map iteration order to pick which commodity. Dropped.

## Motivation

NOTES F1+F2+F3 from `doppio-research/prototypes/doppio-categorize/NOTES.md`. Without these accessors, every `proto::Journal` consumer rewrites:

```rust
let amount = match &posting.amount {
    Some(a) => a,
    None => return None, // or skip, or whatever
};
let dec = amount.by_commodity.get("USD")?;
let value = decimal_from_proto(dec);
```

Real / about-to-be-real consumers writing this dance:
- doppio-cli `register` / `print` / `balance` (post-#113, iterating `proto::Journal` directly)
- doppio-categorize `Index::build_from_proto`
- bb-ledger / bookie at import time (when migrated to proto-based pipelines)
- Future dashboards, charts, summary tools

A multilingual implication: every language binding will want the equivalent method-style accessors on its generated proto class. The canonical recipe documented in #119 should mirror this shape.

## Acceptance

- [ ] `proto::Amount::iter(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_` implemented as inherent impl.
- [ ] `proto::Amount::get(&self, commodity: &str) -> Option<rust_decimal::Decimal>` implemented.
- [ ] `proto::Posting::amounts(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_` implemented (returns empty iterator if `self.amount` is `None`).
- [ ] `proto::Posting::amount_in(&self, commodity: &str) -> Option<rust_decimal::Decimal>` implemented (returns `None` if `self.amount` is `None`).
- [ ] All four methods live in a module adjacent to the prost-generated code, not inside the generated file (so prost regeneration doesn't clobber them).
- [ ] Doc comments on each note: HashMap iteration order is unspecified; `Posting::*` methods handle the `Option<Amount>` wrapper.
- [ ] Tests cover: empty (`amount: None`), empty `by_commodity`, single-commodity hit/miss, multi-commodity all-iter, multi-commodity specific-get hit/miss.
- [ ] doppio-categorize `Index::build_from_proto` simplifies to use these accessors (out of repo, follow-up).

## Out of scope

- **Case (a) helper**: a higher-level "convert and sum to target commodity" helper that takes a Posting (or set of Postings) and a price/conversion context. The shape of the price-context API isn't clear yet — file a separate issue once Phase C clarifies what consumers actually need from `journal.prices`.
- Equivalent `Amount::iter()` / `get()` on `elaboration::Amount` — separate issue (#123, parallel ergonomic gap on the elaboration types). The two type families should land symmetric APIs but track separately.
- Display / Debug improvements on `proto::Decimal` — separate issue (#116).
- Inherent `to_decimal()` on `proto::Decimal` — separate issue (#117). The `iter()` / `get()` / `amounts()` / `amount_in()` methods will use `to_decimal()` internally once that lands.

alevy:
## Summary

- Adds `proto::Amount::iter()` — iterate `(commodity, Decimal)` pairs (order unspecified, per protobuf HashMap semantics)
- Adds `proto::Amount::get(commodity)` — look up one commodity by name
- Adds `proto::Posting::amounts()` — iterate through the `Option<Amount>` wrapper, yielding nothing if `amount` is `None`
- Adds `proto::Posting::amount_in(commodity)` — get a commodity value through the `Option<Amount>` wrapper

All four methods absorb the F1–F3 boilerplate (Option unwrap, HashMap lookup, proto::Decimal reassembly) that every `proto::Journal` consumer was writing by hand. They compose the existing `proto::Decimal::to_decimal()` method added in `proto_ext.rs`.

The design follows the comment thread on #114: a proposed `single_amount()` method was dropped because it conflates three distinct cases (no amount / empty map / multi-commodity) into one `None` and relies on undefined HashMap iteration order.

Both new modules live in sibling files (`proto_accessors.rs`, `proto_ext.rs`) so prost regeneration never clobbers them.

## Test plan

- [ ] `Amount::iter`: empty map, single-commodity, multi-commodity (order-independent assertion)
- [ ] `Amount::get`: hit, miss
- [ ] `Posting::amounts`: `amount: None`, `amount: Some(empty)`, single-commodity, multi-commodity
- [ ] `Posting::amount_in`: `amount: None`, commodity absent, commodity present
- [ ] `cargo fmt --all --check` — clean
- [ ] `cargo clippy --workspace -- -D warnings` — clean
- [ ] `cargo test --workspace` — all pass (173 tests + 12 new)
- [ ] `cargo build --target wasm32-unknown-unknown -p doppio --lib` — clean

Closes #114.

### DIFF
--- a/crates/doppio/src/proto_ext.rs
+++ b/crates/doppio/src/proto_ext.rs
@@ -24,6 +24,45 @@ impl std::fmt::Display for proto::Decimal {
     }
 }
 
+impl proto::Amount {
+    /// Iterate `(commodity, decimal)` pairs in this amount.
+    ///
+    /// Order is unspecified — the underlying `by_commodity` map is a
+    /// `HashMap`, and the protobuf spec does not guarantee map field ordering.
+    pub fn iter(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_ {
+        self.by_commodity
+            .iter()
+            .map(|(k, v)| (k.as_str(), v.to_decimal()))
+    }
+
+    /// Return the decimal value for `commodity`, or `None` if absent.
+    pub fn get(&self, commodity: &str) -> Option<rust_decimal::Decimal> {
+        self.by_commodity
+            .get(commodity)
+            .map(proto::Decimal::to_decimal)
+    }
+}
+
+impl proto::Posting {
+    /// Iterate `(commodity, decimal)` pairs across this posting's amount.
+    ///
+    /// Yields nothing if `self.amount` is `None` or the amount's
+    /// `by_commodity` map is empty.
+    pub fn amounts(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_ {
+        self.amount.iter().flat_map(|a| {
+            a.by_commodity
+                .iter()
+                .map(|(k, v)| (k.as_str(), v.to_decimal()))
+        })
+    }
+
+    /// Return the decimal value for `commodity` in this posting's amount,
+    /// or `None` if the amount field is absent or `commodity` is not present.
+    pub fn amount_in(&self, commodity: &str) -> Option<rust_decimal::Decimal> {
+        self.amount.as_ref()?.get(commodity)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::{decimal_from_proto, proto};
@@ -46,6 +85,16 @@ mod tests {
             mantissa_low: mantissa as u64,
             scale: d.scale(),
         }
+    }
+
+    /// `proto::Decimal` for 7.58 — mantissa 758, scale 2.
+    fn dec_7_58() -> proto::Decimal {
+        make_decimal(0, 758, 2)
+    }
+
+    /// `proto::Decimal` for 1.23 — mantissa 123, scale 2.
+    fn dec_1_23() -> proto::Decimal {
+        make_decimal(0, 123, 2)
     }
 
     // ──────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


alevy:
Force-pushed to descope: removed `proto_ext.rs` (and the `mod proto_ext;` declaration) since that's the canonical scope of #117 / PR #129. Switched the four accessor methods back to calling `crate::decimal_from_proto` directly. The PR is now strictly the `proto::Amount` / `proto::Posting` accessor work for #114 — 2 files changed, +235 / 0.

Pre-push gates re-verified locally: fmt, clippy --workspace -- -D warnings, cargo test --workspace (12 new `proto_accessors` tests pass alongside everything else), and wasm32-unknown-unknown build.

Once #129 (`Decimal::to_decimal()`) merges, a small follow-up can switch the call sites here from `crate::decimal_from_proto(v)` to `v.to_decimal()`. That's a 4-line cleanup, not blocking review of this PR.

alevy:
Rebased on top of post-#129 main and applied the cleanup: the four call sites now use `v.to_decimal()` (and `proto::Decimal::to_decimal` as the function-pointer in the `.map(...)` slot) instead of `crate::decimal_from_proto(v)`. `proto_ext` is now a real dependency of `proto_accessors`, which was the original plan.

Diff stays small: 2 files changed, +234 / 0 (`proto_accessors.rs` new and `mod proto_accessors;` line in lib.rs).

Pre-push gates re-verified locally: fmt, clippy --workspace -- -D warnings, cargo test --workspace (177 passing across the workspace), wasm32-unknown-unknown lib build.

alevy:
Folded the four `proto::Amount` / `proto::Posting` accessors and their tests into `proto_ext.rs` alongside `Decimal::to_decimal()` and the Display impl. `proto_accessors.rs` is deleted and `mod proto_accessors;` removed from lib.rs. Single helpers module now.

Diff: `+208 / 0`, one file (`proto_ext.rs`) modified, plus the lib.rs mod-line cleanup. All four gates re-verified locally.

PR #130, opened 2026-04-28
docs(proto): document Amount map iteration-order non-determinism (closes #118)

### LINKED ISSUE #118: [Phase C] Document HashMap iteration-order non-determinism for `proto::Amount`
## Goal

Document explicitly that `proto::Amount.by_commodity` (and any other `map<...>` field in the schema) has **undefined iteration order** per the protobuf spec. Consumers that need deterministic ordering must sort.

Add a comment block to `proto/doppio.proto` near the `Amount` message; mirror the warning in the Rust-side `proto::Amount` doc comment.

## Motivation

NOTES F3 from `doppio-research/prototypes/doppio-categorize/NOTES.md`. The elaboration-side `Amount` uses `BTreeMap` (deterministic); the proto-side uses `HashMap` (non-deterministic). For consumers that hash, display, or serialize commodities, this is a real divergence.

The doppio-categorize use case (iterating to find a single commodity) doesn't care about order, but consumers doing display/cache-key/text-roundtrip will.

## Acceptance

- [ ] `proto/doppio.proto`: comment on `Amount.by_commodity` explicitly notes undefined iteration order, recommends sorting by commodity symbol for deterministic output.
- [ ] Rust-side `proto::Amount` doc comment (in the module that re-exports/wraps the prost type) carries the same warning.
- [ ] If any other proto message has a `map<...>` field, apply the same treatment.

## Out of scope

- Switching to `BTreeMap` on the Rust side — protobuf's wire-level map is unordered; even if Rust used a sorted container, cross-language consumers can't rely on it. Documentation is the right fix.

alevy:
- Adds an iteration-order warning to `Amount` in `proto/doppio.proto`, noting that `map<string, Decimal>` has unspecified order per the protobuf spec, and showing the canonical sort pattern. The same note calls out `Posting.metadata`, `Transaction.metadata`, `Journal.accounts`, and `Journal.commodities` as equally unordered.
- Adds `Amount::iter` and `Posting::amounts` convenience methods to `crates/doppio/src/proto_ext.rs` with the iteration-order warning and sort recipe in their doc comments.

Closes #118.

### DIFF
--- a/crates/doppio/src/proto_ext.rs
+++ b/crates/doppio/src/proto_ext.rs
@@ -29,6 +29,13 @@ impl proto::Amount {
     ///
     /// Order is unspecified — the underlying `by_commodity` map is a
     /// `HashMap`, and the protobuf spec does not guarantee map field ordering.
+    /// If you need stable, deterministic output (for display, cache keys, or
+    /// text round-trips) sort before consuming:
+    ///
+    /// ```rust,ignore
+    /// let mut pairs: Vec<_> = amount.iter().collect();
+    /// pairs.sort_by_key(|(c, _)| *c);
+    /// ```
     pub fn iter(&self) -> impl Iterator<Item = (&str, rust_decimal::Decimal)> + '_ {
         self.by_commodity
             .iter()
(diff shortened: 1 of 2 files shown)


alevy:
Force-pushed to descope. The B agent's worktree branched off main before #128 merged, so it didn't see the `proto::Amount::iter`, `proto::Posting::amounts`, etc. methods that already exist there — and re-added them with mismatched signatures (`(&str, &proto::Decimal)` instead of `(&str, rust_decimal::Decimal)`).

Rebased on current main and reverted those new `impl` blocks, since the methods already exist with the right signatures. The remaining changes are pure docs:

1. `proto/doppio.proto`: comment block on `message Amount` documenting that protobuf map iteration order is unspecified, with a Rust `sort_by_key` example. Also calls out the same pattern for the other map fields in the schema.
2. `crates/doppio/src/proto_ext.rs`: extended the existing doc comments on `proto::Amount::iter` and `proto::Posting::amounts` with the canonical sort pattern (the `rust,ignore` block from the agent's draft, applied to the methods that actually shipped).

Diff: `+27 / -1`, two files. All gates re-verified locally.

Issue #121, opened 2026-04-28
[Phase C] Decision: deprecate `elaboration::Journal` as public read type


alevy:
## Goal

Once the Phase C convenience accessors (#114-#119) have landed and downstream Rust consumers have migrated (#120), make a deliberate decision about the public visibility of `elaboration::Journal` for v1.0:

1. **Demote to crate-internal**: keep `elaboration::Journal` as the elaboration stage's working type, but no longer expose it as a public read API. Public consumers go through `proto::Journal`. Aggressive — may break consumers we don't know about.
2. **Public-but-discouraged**: keep it public, document `proto::Journal` as the recommended read API, mark `elaboration::Journal` as "stable but not the canonical surface." Conservative — preserves any external consumer not yet identified.
3. **Status quo + dual-API**: keep both fully supported. Punt the decision to v1.1 or later. Safest, but pays the dual-API maintenance cost forever.

This is a v1.0 boundary decision. The decision itself is the deliverable; do not implement it inside this issue.

## Why this is a separate issue

The decision depends on real signal from #120 (migration). If the convenience accessors aren't sufficient, the right answer is option 3 plus more accessors. If migration is clean and no friction surfaces, option 1 or 2 becomes credible.

### Available now (post-Phase-C, 2026-04-28)

- **doppio-categorize migrated cleanly** (commit `0a96ef5` in alevy/doppio-research). The 10-line `.as_ref().map().unwrap_or_default()` dance collapsed to a single `posting.amounts().map(|(_, d)| d).collect()` call. Parity test passes; eval harness produces byte-identical results via `--use-proto`. Updated verdict in `doppio-research/prototypes/doppio-categorize/NOTES.md`: **the proto path is now the same length as the elaboration path, sometimes shorter**. The "5-10% boilerplate gap" estimated in v0.2's NOTES.md is closed.
- **doppio CLI** has been on the proto path since PR #113 with no observed friction.
- **All five F1-F5 friction points** documented in the categorizer's NOTES.md are now closed by Phase C accessors and the Display impl.

### Pending

- **bb-ledger / bookie migrations** (tracked in #120). One real downstream consumer (the categorizer) has migrated cleanly; we need at least one of these to corroborate before this decision crosses from "credible" to "supported by data."
- Whether any new external Rust consumer surfaces between now and the decision point.

## Acceptance

- [ ] At least one of bb-ledger / bookie has migrated to `proto::Journal` (or has a documented blocker).
[...]

PR #132, opened 2026-04-28
Rename proto -> elaboration; demote old elaboration types to crate-private (#121)

alevy:
## Summary

Implements the namespace half of #121. The prost-generated types now live at `doppio::elaboration::*` (the public canonical read API). The previous BTreeMap-based pipeline types are demoted to `pub(crate) mod elaboration_pipeline` — they remain as a transitional intermediate inside the elaborator but are no longer visible to external callers.

## Public API changes (BREAKING)

- `doppio::Journal` is now `doppio::elaboration::Journal` (= former `doppio::proto::Journal`).
- `doppio::elaboration::*` is now the prost-generated namespace (`Journal`, `Transaction`, `Posting`, `Amount`, `Decimal`, etc.).
- `compile()`, `read_dop()`, `write_dop()` all take/return the new `Journal` directly. No more conversion at call sites.
- `read_dop_proto()` removed — now identical to `read_dop()`.
- New: `pub fn elaborate(hir: resolution::HIR) -> Result<elaboration::Journal, ElaborationError>` so external callers (CLI etc.) don't need to reach for the crate-private pipeline types.
- New: `elaboration::Posting::amount(&self) -> &elaboration::Amount` infallible accessor that papers over the proto3 `Option<Amount>` quirk per the #121 discussion (returns a static empty `Amount` fallback if somehow `None`).
- `ElaborationError` and `EvaluationError` re-exported at the crate root.

## Migration recipe for downstream consumers

```rust
// Before
use doppio::elaboration::{Journal, ResolvedTransaction, ResolvedPosting, Amount};
let journal: doppio::Journal = doppio::compile(...)?;
for txn in &journal.transactions {
    for posting in &txn.postings {
        for (commodity, amount) in posting.amount.0.iter() {
            // ... use amount: Decimal
        }
    }
}

// After
use doppio::elaboration::{Journal, Transaction, Posting, Amount};
let journal: doppio::Journal = doppio::compile(...)?;
for txn in &journal.transactions {
    for posting in &txn.postings {
        for (commodity, amount) in posting.amounts() {
            // ... use amount: Decimal (Phase C accessor)
        }
    }
}
```

The old `elaboration::ResolvedTransaction` / `ResolvedPosting` / `Amount` types are gone from the public surface. The replacements (`elaboration::Transaction` etc.) are proto-shaped (HashMap maps, `Option<Amount>` field, etc.) but Phase C accessors (`posting.amounts()`, `posting.amount_in("$")`, `posting.amount()`) cover all the real use patterns ergonomically.

## Internal

- `crates/doppio/src/elaboration.rs` → renamed to `elaboration_pipeline.rs` with module visibility `pub(crate)`. The elaborator's internal `TryFrom<HIR>` impl still produces these types; conversion to the public `elaboration::Journal` happens in the new `elaborate()` helper at the API boundary.
- `crates/doppio/src/proto_ext.rs` → renamed to `elaboration_ext.rs`, contents updated for the new namespace. New `Posting::amount()` accessor added.
- All `crate::proto::*` references updated to `crate::elaboration::*`.

## Verified

- All 177 workspace tests pass
- `cargo fmt --all --check` clean
- `cargo clippy --workspace -- -D warnings` clean
- `cargo build --target wasm32-unknown-unknown -p doppio --lib` clean

## Out of scope

- Refactoring the elaborator's internal types to use proto-shaped representations directly (i.e. eliminating the crate-private pipeline types). Tracked separately because it's optimization, not API change.
- Downstream consumer updates (`alevy/doppio-research/prototypes/doppio-categorize`, bb-ledger PoC, bookie PoC). Companion commits in the doppio-research repo land separately. The PoC migrations from earlier in this sprint will need a small ripple to track the rename.
- CHANGELOG entry — will add as part of v0.4 cut.

Closes the namespace half of #121. The "decide demote vs deprecate" half of #121 is also satisfied: the old elaboration types are demoted (option A1) and the proto types take over the namespace (option B2).

### DIFF
--- a/crates/doppio-cli/examples/list_accounts.rs
+++ b/crates/doppio-cli/examples/list_accounts.rs
@@ -41,15 +41,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Accumulate per-account, per-commodity balances by walking every posting
     // in every transaction.  Running state is no longer stored on individual
     // transactions, so we compute it here from scratch.
-    let mut balances: BTreeMap<&String, BTreeMap<&String, Decimal>> = BTreeMap::new();
+    let mut balances: BTreeMap<&String, BTreeMap<String, Decimal>> = BTreeMap::new();
     for txn in journal.transactions.iter() {
         for posting in txn.postings.iter() {
-            for (commodity, amount) in posting.amount.0.iter() {
+            for (commodity, amount) in posting.amounts() {
                 *(balances
                     .entry(&posting.account)
                     .or_default()
-                    .entry(commodity)
-                    .or_default()) += *amount;
+                    .entry(commodity.to_string())
+                    .or_default()) += amount;
             }
         }
     }
--- a/crates/doppio-cli/src/bin/timings.rs
+++ b/crates/doppio-cli/src/bin/timings.rs
@@ -43,7 +43,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // --- elaboration ---
     // Expression evaluation, transaction balancing, account registration.
     let t3 = Instant::now();
-    let journal: doppio::elaboration::Journal = hir.try_into()?;
+    let journal: doppio::elaboration_pipeline::Journal = hir.try_into()?;
     eprintln!("elaboration: {:>8.3}s", t3.elapsed().as_secs_f64());
 
     // --- serialize ---
--- a/crates/doppio-cli/src/bin/timings.rs
+++ b/crates/doppio-cli/src/bin/timings.rs
@@ -43,7 +43,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // --- elaboration ---
     // Expression evaluation, transaction balancing, account registration.
     let t3 = Instant::now();
-    let journal: doppio::elaboration_pipeline::Journal = hir.try_into()?;
+    let journal: doppio::elaboration::Journal = doppio::elaborate(hir)?;
     eprintln!("elaboration: {:>8.3}s", t3.elapsed().as_secs_f64());
 
     // --- serialize ---
--- a/crates/doppio-cli/src/main.rs
+++ b/crates/doppio-cli/src/main.rs
@@ -315,7 +315,7 @@ impl JournalFilter {
 /// Returns `None` if the commodity has no declared format.
 fn commodity_format<'a>(
     commodity: &str,
-    commodities: &'a std::collections::HashMap<String, doppio::elaboration::CommodityProperties>,
+    commodities: &'a std::collections::BTreeMap<String, doppio::elaboration::CommodityProperties>,
 ) -> Option<&'a str> {
     commodities.get(commodity).and_then(|p| p.format.as_deref())
 }
--- a/crates/doppio-cli/src/main.rs
+++ b/crates/doppio-cli/src/main.rs
@@ -152,7 +152,7 @@ impl OutputFormat {
     }
 }
 
-/// Load a [`doppio::proto::Journal`] from either a compiled `.dop` file or a
+/// Load a [`doppio::elaboration::Journal`] from either a compiled `.dop` file or a
 /// raw source file, without performing the `proto::Journal → elaboration::Journal`
 /// reverse conversion.
 ///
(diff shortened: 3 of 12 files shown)


alevy (commented):
Reviewed. Approved + suggested doc cleanups.

## What landed in `0f1ace1` is genuinely good

Two-line config change + four-line simplification, but it has outsized effects:

1. **`prost_build::Config::new().btree_map(["."]).compile_protos(...)`** — tells prost to generate `BTreeMap` instead of `HashMap` for every map field in the schema (`"."` glob matches everything). Wire format is unchanged (the on-disk bytes are the same key-value pairs); only the consumer-side iteration shape changes. Old `.dop` files still decode correctly.

2. **`Posting::amount()` static fallback** can now use `BTreeMap::new()` (a `const fn`) instead of `OnceLock`, because the empty Amount is constructible at compile time. Cleaner code path, no thread-local lazy-init machinery, works in environments that are unhappy with `OnceLock` (`no_std`, certain WASM runtimes).

Plus, indirectly:

3. **The Option II elaborator-internal refactor I previously argued was risky becomes nearly a no-op.** With both `elaboration_pipeline::Amount(BTreeMap)` and `proto::Amount { by_commodity: BTreeMap }` shape-identical at the BTreeMap layer, the elaborator could be rewritten to produce proto types directly with very small changes — the BTreeMap accumulation logic transfers wholesale. Worth filing as a follow-up; the rationale I gave for deferring no longer holds at the same strength. Not blocking #132.

## Stale doc comments to update before merge

Three places still say HashMap or "order is unspecified" but are now wrong at the doppio Rust binding layer:

1. **`crates/doppio/src/elaboration_ext.rs:30-31`** (`Amount::iter` docstring):
   > Order is unspecified — the underlying `by_commodity` map is a `HashMap`, and the protobuf spec does not guarantee map field ordering.

   Now: order IS deterministic (BTreeMap iterates sorted by commodity symbol). The example `sort_by_key` block can be removed.

2. **`crates/doppio/src/elaboration_ext.rs:78`** (`Posting::amounts` docstring): same — "Order is unspecified" line points at `Amount::iter` for the canonical pattern. Both should say something like "yields entries in commodity-symbol order (BTreeMap)".

3. **`proto/doppio.proto:96`** (the multilingual iteration-order block from #118):
   > and prost generates a `HashMap<String, Decimal>` on the Rust side.

   Now: prost generates a `BTreeMap`. The protobuf-spec point still stands for **other languages' bindings** (the spec doesn't guarantee order, so language X's binding may iterate however it wants), but the doppio-Rust-binding clause is wrong. Worth rewording so it's clear which guarantees apply at which layer.

## Tests now have redundant sorts

`elaboration_ext.rs:tests::amount_iter_multi_commodity_all_present` and `posting_amounts_multi_commodity` do `pairs.sort_by_key(|(c, _)| *c)` before asserting. Harmless but redundant. Optional cleanup.

## Approval

LGTM modulo the three doc fixes above. I can land the doc fixes as a follow-up commit on this branch if you'd like. The architectural change is sound.

## Implication for #121's framing

The user-visible spec for #121 was "replace the elaboration namespace with the protocol buffers," and after this PR plus the BTreeMap config, the proto types are now *more* like the old elaboration types than I was treating them. The transitional intermediate (`elaboration_pipeline`) is now genuinely just a temporary name for what's effectively the same data shape — strong support for the Option II follow-up.

alevy (commented):
Quick approval review — see comment for details. LGTM modulo three stale doc comments to update (`Amount::iter` and `Posting::amounts` say "Order is unspecified, HashMap...", and `proto/doppio.proto:96` says "prost generates a HashMap"). Happy to land those as a follow-up commit on this branch.

alevy (commented):
Reviewed — all gates green locally. The btree_map config change is great.

alevy:
## Full review

The btree_map config change is a clean, leveraged improvement. Two-line config + four-line simplification, but it cascades nicely:

1. **Wire format unchanged** — protobuf map<> wire bytes are the same regardless of consumer-side container choice. Old `.dop` files decode fine.

2. **`Posting::amount()` static fallback** can now use `const fn BTreeMap::new()` instead of `OnceLock`. No thread-local lazy-init machinery, friendlier for environments unhappy with `OnceLock` (`no_std`, certain WASM runtimes).

3. **Strengthens the Option II elaborator-internal refactor case I previously deferred.** With both `elaboration_pipeline::Amount(BTreeMap)` and `proto::Amount { by_commodity: BTreeMap }` now shape-identical, the elaborator could produce proto types directly with very small changes — the BTreeMap accumulation logic transfers wholesale. Worth filing as a follow-up; my earlier risk argument no longer holds. Not blocking #132.

## Stale doc comments to update

Three places still describe HashMap / unspecified-order semantics:

- `crates/doppio/src/elaboration_ext.rs:30-31` (`Amount::iter` docstring): "Order is unspecified — the underlying `by_commodity` map is a `HashMap`..."
- `crates/doppio/src/elaboration_ext.rs:78` (`Posting::amounts` docstring): "Order is unspecified — see `Amount::iter` ..."
- `proto/doppio.proto:96` (multilingual iteration-order block from #118): "and prost generates a `HashMap<String, Decimal>` on the Rust side."

The proto-file comment is interesting: the protobuf spec point ("map order is unspecified per the spec") still stands for **other languages' bindings**, but the doppio-Rust-binding claim is now wrong. Worth rewording to make the layered guarantees clear.

Going to land these as a follow-up commit on this branch — and clean up the now-redundant `sort_by_key` calls in the test fixtures while I'm in there.

Issue #111, opened 2026-04-27
Optimize dop balance/register/print: iterate proto::Journal directly, skip reverse conversion


alevy:
## Background

PR #109 introduced a `proto::Journal ↔ elaboration::Journal` conversion on every read/write of `.dop` files. Empirical measurements (criterion + wall-clock) show this conversion is a **non-trivial fraction of `dop balance` load time**:

| Stage | 10k tx | 100k tx |
|---|---:|---:|
| Prost decode | ~24% | ~26% |
| Reverse conversion (`proto::Journal → elaboration::Journal`) | 6.2% | 19.9% |
| **Decode + reverse total** | **~30%** | **~46%** |
| File I/O + inflate + iterate balance | ~70% | ~54% |

At 100k transactions, **roughly half** of `dop balance` wall-clock is spent decoding the protobuf and translating into `elaboration::Journal`. The reverse conversion is **0.7 μs / txn** and allocates a fresh `Journal` with `BTreeMap`s, `String`s for accounts/payees/descriptions, `Amount` wrappers around per-posting commodity maps, and clones every `metadata` map — even for transactions a downstream filter would discard.

## Proposal

Have the CLI commands that **only read and iterate** (`balance`, `register`, `print`, plus likely `stats`, `accounts`, `commodities`) operate directly on the prost-decoded `proto::Journal`, skipping the reverse conversion.

The library API contract — `doppio::Journal` (= `elaboration::Journal`) as the public Rust type returned by `read_dop` — **does not change**. Downstream Rust consumers (bb-ledger, etc.) keep what they have.

The CLI just gets a parallel cheaper path:

```rust
// crates/doppio-cli/src/main.rs (sketch)
fn load_journal_proto(path: &Path) -> Result<doppio::proto::Journal, _> {
    // Read header, inflate, prost-decode — skip the From conversion.
}
```

Then refactor `balance` / `register` / `print` to use a small adapter that walks `proto::Journal::transactions` and constructs whatever running state it needs from the proto types. Decimal arithmetic still goes through `rust_decimal` (one conversion per posting touched, not per posting in the whole journal).

## Estimated win

`dop balance` 100k tx: **~73 ms / 370 ms ≈ 20% wall-clock reduction**. Prost decode itself stays.

## Out of scope

- Streaming-decoding directly into the balance loop (a custom decoder, much more code, probably not worth it until someone hits a multi-million-transaction journal).
[...]

PR #134, opened 2026-04-28
Elaborator emits elaboration::Journal directly; delete transitional types (closes #133)

### LINKED ISSUE #133: [Phase C] Make elaborator emit elaboration::Journal directly; delete elaboration_pipeline transitional types
## Goal

After PR #132 lands, doppio's public surface uses the prost-generated `elaboration::*` types, but the elaborator (`crates/doppio/src/elaboration_pipeline.rs::TryFrom<resolution::HIR> for Journal`) still constructs the old BTreeMap-based wrapper types and converts at the API boundary via `(&pipeline_journal).into()`. This issue eliminates that transitional intermediate: rewrite the elaborator's `TryFrom` impl to construct `elaboration::Journal` (= the proto-generated type) directly.

## Why this is now feasible (and was previously deferred)

When I sketched the deferral in #121's discussion, the elaborator's `Amount(BTreeMap)` and `RunningState`-style helpers depended on deterministic key ordering, while `proto::Amount.by_commodity` was a `HashMap`. That delta was the load-bearing reason for the boundary-conversion approach.

PR #132's `btree_map(["."])` config change in `build.rs` flipped prost to generate `BTreeMap` for every map field. The proto-side and pipeline-side types are now structurally identical at the BTreeMap layer — the conversion in `From<&pipeline_journal::Journal> for elaboration::Journal` is now just struct-shape rearrangement (Decimal → proto::Decimal via existing `decimal_to_proto`; TransactionState enum → i32 via existing `state_to_proto`; `Amount(BTreeMap<...>)` → `Some(proto::Amount { by_commodity: BTreeMap<...> })`). No iteration order, no algorithmic concerns, no risk to balance-assertion / null-posting machinery.

## PR structure (review-driven)

To make the diff easy to verify as **algorithmically unchanged** rather than a rewrite, structure the PR as separate commits in this order:

### Commit 1 — `try_from` body change (the load-bearing commit)

Change `impl TryFrom<resolution::HIR>` to produce `elaboration::Journal` instead of `pipeline_journal::Journal`. The diff should read as **field-by-field struct constructor swaps**, not new logic:

- `transactions.push(ResolvedTransaction { ... })` → `transactions.push(elaboration::Transaction { ... })`. Field shapes parallel: `date` (i32) stays, `state: TransactionState::Cleared` → `state: state_to_proto(&TransactionState::Cleared)`, `postings` field with the same Vec construction, etc.
- `ResolvedPosting { account, payee, amount: Amount(map), ... }` → `elaboration::Posting { account, payee, amount: Some(elaboration::Amount { by_commodity: map }), state: state_to_proto(...), ... }`.
- `Amount(by_commodity)` accumulator stays as a local helper type *during* per-transaction balancing — only the OUTPUT values become proto-shaped. The map type is BTreeMap on both sides post-#132, so the accumulator's `entry().and_modify().or_insert()` arithmetic transfers wholesale.
- Decimal values get `decimal_to_proto(...)` applied at the point they enter a proto::Amount or proto::HistoricalPrice.

Also in this commit: update `pub fn elaborate(hir)` to remove the now-redundant boundary conversion (`Ok(hir.try_into()?)` instead of pipeline-then-convert).

The expected diff: ~150 lines of struct-constructor swaps. Anyone reading this commit should be able to verify it's **purely a different way to populate the output**, with no branching changes, no new logic, no algorithmic shifts.

### Commit 2 — Remove dead conversion impl

Delete `impl From<&pipeline_journal::Journal> for elaboration::Journal` (and the owned `From<elaboration_pipeline::Journal>` if it exists) in `crates/doppio/src/lib.rs`. After commit 1 these are unused.

### Commit 3 — Delete transitional output types

In `elaboration_pipeline.rs`, delete the now-unused `pub(crate)` output struct types: `Journal`, `ResolvedTransaction`, `ResolvedPosting`, `AccountProperties`, `CommodityProperties`, `HistoricalPrice`, the standalone `Amount` type (if no longer referenced), `TransactionState` enum (if no longer referenced).

**Keep**:
- Internal accumulation helpers used by the elaborator's algorithm: `AccountBalances`, `RunningState`, possibly `Amount` if it's still useful as a local accumulator type during transaction balancing.
- Error types: `ElaborationError`, `EvaluationError` (re-exported at the crate root).

### Commit 4 (optional) — File rename + docs

If the file is now exclusively elaboration-algorithm code (no transitional types), consider renaming `elaboration_pipeline.rs` → `elaborator.rs` to match the new role. Update `mod` declaration in lib.rs. Update doc comments referencing "pipeline intermediate."

## Acceptance

- [ ] All 177 workspace tests still pass after Commit 1 (algorithmic no-op verification).
- [ ] All four pre-push gates clean (`cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo test --workspace`, `cargo build --target wasm32-unknown-unknown -p doppio --lib`).
- [ ] Commit 1's diff is reviewable as struct-constructor changes; no logic restructuring sneaks in.
- [ ] After Commit 3, `crates/doppio/src/elaboration_pipeline.rs` (or its renamed equivalent) contains only the elaborator algorithm + internal helpers + error types. No public output struct types.
- [ ] Internal performance: `compile()` and `read_dop` no longer pay the per-load boundary conversion cost (~50ms per 100k transactions per the #111 measurement). Spot-check via the existing benchmark harness if convenient.

## Out of scope

- Refactoring the elaborator's *algorithm* (balance assertions, null-posting fill, expression evaluation). The whole point of this issue is to NOT change the algorithm — only the output type.
- Public API changes beyond `elaborate()`'s body simplification. Signatures stay the same.
- Downstream consumer ripple (doppio-categorize, bb-ledger PoC, bookie PoC). Those don't observe the internal change.

## Related

- Closes the "internal" half of #121 (the namespace half is closed by #132).
- Enables a CHANGELOG entry for v0.4 / v1.0-rc noting "elaboration emits proto types directly; per-load boundary conversion eliminated."

alevy:
## Summary

Per #133. The elaborator's `TryFrom<resolution::HIR>` impl now constructs `elaboration::Journal` (= the prost-generated proto-shape type) directly at every output site. The transitional `elaboration_pipeline` types and the per-load boundary conversion are deleted.

Net diff: **+89 / -423** across two files.

### Commit 1: `Refactor: TryFrom<resolution::HIR> emits elaboration::Journal directly` (+77 / -114)

Field-by-field struct constructor swaps in the elaborator's `try_from`. Reads as **purely a different way to populate the output**, not as new logic — no branching changes, no algorithmic shifts.

The substantive changes:
- `ResolvedTransaction { ... }` → `elaboration::Transaction { ... }` with `state: state_to_proto(&...)`.
- `ResolvedPosting { ... }` → `elaboration::Posting { ... }` with `amount: Some(elaboration::Amount { by_commodity })`. `by_commodity` built by `decimal_to_proto`'ing each `Decimal` value.
- `HistoricalPrice { ... }` → `elaboration::HistoricalPrice { ... }` with `price: Some(decimal_to_proto(...))`.
- `AccountProperties { ... }` / `CommodityProperties { ... }` / final `Journal { ... }` — direct namespace swap (same field shapes since `by_commodity` is BTreeMap on both sides per #132's btree_map config).

Inner per-commodity loops over the just-constructed proto Postings now use `posting.amounts()` (the Phase C accessor) instead of `posting.amount.0.iter()`, since the field shape changed from `Amount(BTreeMap)` to `Option<elaboration::Amount>`. Same data, different access syntax.

Internal accumulators (`Amount(BTreeMap)` for transaction balancing, `RunningState`, `AccountBalances`) are **unchanged** — they're still local helpers used by the elaboration algorithm.

Also in this commit:
- `pub fn elaborate(hir)` simplifies to `elaboration::Journal::try_from(hir)`.
- `pub fn eval_transaction` now returns `elaboration::Transaction`.
- 76 elaborator tests + 8 `eval_transaction` tests updated mechanically: `posting.amount.0.get(c).copied()` → `posting.amount_in(c)`, `matches!(t.state, TransactionState::Cleared)` → `t.state == elaboration::TransactionState::Cleared as i32`, `price.price` → `price.price.as_ref().unwrap().to_decimal()`.
- `proto_from_journal_tests` module removed (it tested the `From` impls being deleted in commit 2; equivalent coverage from CLI's `dop_format` round-trip tests).

### Commit 2: `Delete dead transitional types and conversion impls` (0 / -291)

After commit 1 the transitional types in `elaboration_pipeline.rs` are unreferenced and the `From` conversion impls in `lib.rs` are dead. Deleted:

- In `lib.rs`: `From<&pipeline::Journal> for elaboration::Journal`, `From<pipeline::Journal> for elaboration::Journal` (owned), `From<elaboration::Journal> for pipeline::Journal`, `state_from_proto`.
- In `elaboration_pipeline.rs`: `Journal`, `CommodityProperties`, `HistoricalPrice`, `AccountProperties`, `ResolvedTransaction`, `ResolvedPosting`.

Kept (still used internally by the elaborator):
- `Amount(BTreeMap<...>)` — transaction-balance accumulator.
- `Commodity` type alias (used by `Amount`).
- `TransactionState` enum + `From<ast::TransactionState>` impl — fed to `state_to_proto` via `posting.state.into()`.
- `AccountBalances`, `RunningState` — internal helpers.
- `ElaborationError`, `EvaluationError` — re-exported at the crate root.

This commit had to bundle the type deletions with the impl deletions because `clippy --workspace -- -D warnings` rejects dead-code warnings. Each individual deletion is verifiable as no-longer-referenced.

### Commit 3: `Rename elaboration_pipeline -> elaborator` (+13 / -19)

The module no longer holds transitional output types — only the elaboration algorithm and its internal helpers. "pipeline" is no longer apt; "elaborator" matches its actual role.

## Acceptance check

- [x] All 175 workspace tests pass after Commit 1 alone (algorithmic-no-op verification — 2 tests removed from the original 177 because they tested the `From` impl being deleted in commit 2; equivalent coverage from `dop_format` integration tests).
- [x] All four pre-push gates clean on each commit: `cargo fmt --all --check`, `cargo clippy --workspace -- -D warnings`, `cargo test --workspace`, `cargo build --target wasm32-unknown-unknown -p doppio --lib`.
- [x] Commit 1's diff reads as struct-constructor swaps; no logic restructuring.
- [x] After Commit 2, no public output struct types remain in the elaborator module.
- [x] Per-load boundary conversion eliminated (was ~50ms per 100k transactions per #111's measurement).

## Out of scope

- Renaming or removing `elaborator::Amount`, `elaborator::TransactionState`, etc. They're still used as internal helpers; if a future PR refactors the elaborator's accumulator data structures, those types could go too.
- Public API changes beyond `elaborate()`'s body simplification. Signatures are stable.
- Downstream consumer ripple — categorizer prototype + bb-ledger / bookie PoC branches need a small follow-up to adapt to the rename.
- CHANGELOG entry — will add as part of v0.4 / v1.0-rc cut.

Closes #133.


alevy:
Bench fixes (force-pushed onto commit 3):

- `benches/pipeline.rs:40` — `doppio::elaboration_pipeline::Journal` → `doppio::elaboration::Journal` (the public canonical read type, what `hir.try_into()` now produces).
- `benches/queries.rs:25` — `posting.amount.0.iter()` → `posting.amounts()` (Phase C accessor; same data, current shape).

Verified locally with `cargo bench --no-run --workspace` — all four bench binaries (parse, pipeline, queries, serial) compile clean. Was missed in the original PR sweep because the workspace test+clippy gates don't include benches by default; `--workspace` doesn't recurse into `[[bench]]` targets without `--no-run`. Adding the bench-build to the standing pre-push checklist is a small follow-up.

PR #155, opened 2026-04-29
feat: --exchange flag + Journal price-lookup helper for FX conversion (closes #141)

### LINKED ISSUE #141: [Phase D] FX conversion: P-directive consumption in balance/register reports
## Gap

doppio parses `P <date> <commodity> <price>` directives and stores
them in `elaboration::Journal.prices`, but no report consumes them.
ledger-cli supports:

```sh
ledger balance --exchange USD            # convert all balances to USD using the latest P quotes
ledger balance -X USD --end 2024-01-01   # convert using P quotes as of a given date
```

Per `docs/SUPPORTED_FEATURES.md`: \"P historical price directive — Parsed
and stored. Rendering / FX conversion not yet applied.\"

## Scope

1. Library API: add a price-lookup helper on `elaboration::Journal`
   that, given `(commodity, target_commodity, optional as_of_date)`,
   returns the relevant historical price. Method-style on the proto
   type, consistent with other Phase C accessors.
2. CLI: add `--exchange COMMODITY` / `-X COMMODITY` to `dop balance`
   and `dop register`. Converts every commodity in the report to the
   target via the journal's `P` price chain, using the latest quote
   at or before the report's as-of date.
3. Multi-hop conversion: if a direct quote is absent (USD→EUR has no
   `P`, but USD→GBP and GBP→EUR exist), follow the chain. Document
   the resolution algorithm in the help text.
4. Missing-quote behaviour: if no path exists, surface the
   commodity unconverted with a stderr warning rather than crashing.
5. Tests: lift `fx_conversion_p_directive.ledger` from #138 and
   fill in assertions.

Out of scope:

- Live market-price feeds (explicitly out per `docs/SUPPORTED_FEATURES.md`)
- `commodity nomarket` is a parse-time flag; this issue treats it as
  \"skip this commodity for FX conversion\" since the field already
  exists.

## Acceptance

- `dop balance --exchange USD` produces USD-converted balances
- The FX parity fixture from #138 passes
- `docs/SUPPORTED_FEATURES.md` flips the FX-rendering note to ✅

alevy:
## Summary

- Adds `Journal::price_at(from, to, as_of)` to `elaboration_ext.rs`: BFS over the `P`-directive price graph with as_of cutoff, most-recent-quote-wins semantics, multi-hop chains, and automatic inverse traversal.
- Wires `--exchange COMMODITY` / `-X COMMODITY` into both `dop balance` and `dop register`; unconvertible commodities warn to stderr and fall back gracefully.
- Lifts the `fx_conversion_p_directive` parity test out of `#[ignore]` and adds 8 unit tests in `elaboration_ext::tests` (direct, inverse, two-hop, no-path, as_of filtering).
- Updates `docs/SUPPORTED_FEATURES.md`: flips the P-directive row note and adds a CLI row for `--exchange`.

## What changed

| File | Change |
|---|---|
| `crates/doppio/src/elaboration_ext.rs` | `impl elaboration::Journal { pub fn price_at(...) }` + unit tests |
| `crates/doppio-cli/src/main.rs` | `--exchange`/`-X` on `balance` and `register`; `maybe_convert_amount` helper |
| `crates/doppio/tests/parity.rs` | `fx_conversion_p_directive` un-ignored, assertions filled in |
| `docs/SUPPORTED_FEATURES.md` | P-directive row + new CLI row |

## Test plan

- [ ] `cargo fmt --all --check` — clean
- [ ] `cargo clippy --workspace -- -D warnings` — clean
- [ ] `cargo test --workspace` — 37 parity tests pass (was 36; `fx_conversion_p_directive` now green); 8 new unit tests pass
- [ ] `cargo bench --no-run --workspace` — compiles
- [ ] `cargo build --target wasm32-unknown-unknown -p doppio --lib` — compiles

Closes #141

### DIFF
--- a/crates/doppio/src/elaboration_ext.rs
+++ b/crates/doppio/src/elaboration_ext.rs
@@ -3,6 +3,7 @@
 //! `OUT_DIR` via the `build.rs` `include!`) doesn't clobber these impls.
 
 use crate::elaboration;
+use std::collections::{BTreeMap, HashMap, VecDeque};
 
 impl elaboration::Decimal {
     /// Reconstruct a [`rust_decimal::Decimal`] from this proto-encoded value.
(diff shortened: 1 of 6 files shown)


crates/doppio/src/elaboration_ext.rs
```diff
@@ -118,6 +119,120 @@ impl elaboration::HistoricalPrice {
...
+    ///
+    /// Conversion uses the report's `--end` date as the as-of cutoff, or the
+    /// latest available quote if `--end` is not specified. ledger-cli converts
+    /// per-posting using the transaction's own date by default; this
+    /// implementation uses a single uniform as-of for all postings, which is
+    /// simpler but means historical reports without `--end` will use
+    /// anachronistically recent rates.
+    ///
+    /// Returns `None` if no conversion path exists.
+    pub fn price_at(
```
alevy: This is a bit of a bummer. It means that computing conversion rates at a particular time: something that is a bit subtle and slightly different algorithms can result in meaningfully different results, is deferred to each user of a dop file/object, meaning it needs to be re-implemented in each language.

In an _ideal_ world, the exchange rates would be computed during elaboration and be present in the compiled data-structure so that `price_at` in each language is a much simpler lookup in the compiled journal (perhaps just backward search of prices to find the most recent pairing before a certain date).

Maybe that's too complicated for this PR, and we should file an issue to try doing that.

alevy: It's not exactly a "price", it's an exchange rate. Maybe `exchange_rate_at`?

alevy: Done — renamed to `exchange_rate_at` throughout (method definition, all unit tests and test names, CLI doc comment, call site in `maybe_convert_amount`, parity test, and `SUPPORTED_FEATURES.md`) in 5a4a6dc.

alevy: Agreed — tracked as #158 to bake exchange rates into elaboration so downstream consumers don't need to re-implement the BFS themselves. Out of scope for this PR.

crates/doppio/src/elaboration_ext.rs
```diff
@@ -118,6 +119,120 @@ impl elaboration::HistoricalPrice {
...
+            if price_val == Decimal::ZERO {
+                continue;
+            }
+            let date = hp.date_naive();
+            if let Some(cutoff) = as_of
+                && date > cutoff
+            {
+                continue;
+            }
+
```
alevy: Try implementing this in terms of filter_maps and/or maps on the `self.prices` iterator, rather than explicit if/continue blocks. Ultimately, we want to iterate over the `(from, to, price_val, date)` for each price entry with a non-zero price and a date before the cutoff. Structuring it that way seems like it might make the algorithm more clear to understand.

alevy: Done — refactored in 7993d4f. The adjacency-build now filters first via `self.prices.iter().filter_map(...)` yielding `(from, to, rate, date)` tuples for non-zero quotes within the as_of cutoff, then the loop body is just two `upsert` calls (forward + inverse) sharing a common closure.

Issue #151, opened 2026-04-29
[Web GUI] JS-side .dop loader via published .proto schema


alevy:
## Goal

The runtime app's primary `.dop` loading path: load a `.dop` file via a native JS/TypeScript protobuf decoder, with **zero Rust/WASM dependency at runtime**. This is the central architectural validator of the format-as-API claim: any non-Rust language consumer can read `.dop` files via the published `proto/doppio.proto` schema, no special tooling required.

Depends on #148 (bootstrap). Complements #149 (the WASM shim path for source files).

## Scope

- Generate TypeScript types from `proto/doppio.proto` at build time. Pick a generator (recommendations: `protobuf-ts`, `ts-proto`, or `@bufbuild/protobuf` — pick one that gives idiomatic TS interfaces and supports `BigInt` for `int64`/`sint64` fields. Document the choice in a comment.).
- Implement `decimalFromProto(p)` per the recipe in the proto file's top comment (uses `BigInt` for the 128-bit mantissa reassembly; converts to `decimal.js` or similar at the boundary).
- Implement `dopReader(buf: Uint8Array): Journal` that:
  - Reads the 8-byte `.dop` header (magic + version + compression byte + reserved); rejects invalid headers with a clear error
  - Inflates via `pako` if the compression byte == 1
  - Decodes the body as `proto::Journal` via the generated TS types
- Render the same Vue views from #150 driven by the JS-decoded Journal — proves the same view code can consume both the WASM-compiled and the protobuf-loaded shapes
- A toggle (or just-let-the-user-drop-either-format) in the UI: drop a `.ledger` → goes through #149's WASM shim; drop a `.dop` → goes through this issue's JS path. The Journal in both cases is structurally identical for the views to consume.

## Acceptance

- A `.dop` produced by `dop compile` loads via the JS path with no WASM in the runtime bundle for that code path
- The same Vue views (#150) render identically whether driven by the WASM source-parser path or the JS-protobuf path
- This is the published-spec validator for `proto/doppio.proto` — surfaces any helper-method gaps the JS recipe needs that the proto top-comment doesn't already cover (file as docs follow-ups against the proto schema)
- The runtime bundle for the `.dop`-only path can be produced **without** the WASM artifact (e.g. via a `npm run build:no-wasm` variant), proving the JS path is genuinely standalone

PR #164, opened 2026-04-29
refactor(fx): drop BFS chaining; direct + inverse only (closes #158)

### LINKED ISSUE #158: [Phase D follow-up] Bake exchange rates into elaboration so cross-language consumers don't re-implement BFS
PR #155 ships \`Journal::exchange_rate_at(from, to, as_of)\` as a query-time helper that walks the \`P\` chain via BFS, with multi-hop, inverse, and most-recent-eligible logic. The user (alevy) flagged on review that this defeats the format-as-API goal:

> This is a bit of a bummer. It means that computing conversion rates at a particular time: something that is a bit subtle and slightly different algorithms can result in meaningfully different results, is deferred to each user of a dop file/object, meaning it needs to be re-implemented in each language.
>
> In an _ideal_ world, the exchange rates would be computed during elaboration and be present in the compiled data-structure so that \`price_at\` in each language is a much simpler lookup in the compiled journal (perhaps just backward search of prices to find the most recent pairing before a certain date).

## Goal

Move FX-conversion-rate resolution out of query-time and into the elaboration stage. The compiled \`elaboration::Journal\` should carry resolved per-pair, per-date exchange rates, so any non-Rust consumer reading the \`.dop\` file performs at most a backward search through a sorted list of (date, from, to, rate) tuples.

## Approach sketch (open for discussion)

1. **At elaboration time**: walk \`journal.prices\` and produce the closure of all (from, to) pairs reachable via BFS at each P-directive's date. Insert inverse edges and multi-hop combinations as derived rates with \`source: derived\` provenance.
2. **Schema**: extend \`proto::Journal\` with an additional \`resolved_rates: repeated ResolvedRate\` field. \`ResolvedRate { date: sint32, from: string, to: string, rate: Decimal, source: enum { DECLARED, INVERSE, DERIVED_VIA: string } }\` — provenance is for downstream auditability.
3. **At query time**: \`exchange_rate_at\` becomes a sorted-list backward search rather than a graph BFS. Identical fast-path for direct quotes; multi-hop and inverse cases also reduce to the same lookup.
4. **Cross-language**: the format-as-API contract becomes \"backward-search through the resolved rates list\" — implementable in 5-10 lines in any language.
5. **Bonus**: the resolved rates can be inspected, displayed, or audited via a new CLI subcommand (\`dop rates\`) that simply prints them.

## Tradeoffs

- **Cost**: elaboration-time CPU + .dop file size grow with the closure of (commodity × date × inverse × hop). Real-world journals usually have a small commodity set, so the closure is tractable; benchmark on bb-org and household corpora before committing.
- **Determinism**: all consumers see exactly the same rate at exactly the same date. Today, language-specific BFS implementations could pick different rates when multiple shortest paths exist (PR #155 documents alphabetical tie-breaking; not all consumers will mirror it).
- **Backwards-compat**: old \`.dop\` files won't have the resolved-rates field; readers fall back to the per-pair-only direct-quote path. Or: add a new \`exchange_rate_at\` semantics version flag in the .dop header.

## Out of scope for this issue

- The existing \`exchange_rate_at\` query-time helper stays for now as a fallback for old .dop files and for tests; removing it is a separate cleanup.
- Per-posting-date FX (rate at the transaction's own date) vs. per-report-as-of FX is a separate UX question that should be settled when this lands.

Tracking from PR #155 review comment: <https://github.com/alevy/doppio/pull/155#discussion_r3162365398>

alevy:
## Summary

Replace `Journal::exchange_rate_at`'s unbounded BFS-with-inverse-edges implementation with a linear scan that returns the most-recent direct or inverse quote for the requested pair. **No chaining through intermediate commodities.** This aligns doppio with Beancount's "non-transitive" design and diverges (deliberately) from hledger/ledger-cli, which chain.

The PR closes #158. The decision is captured in detail in two comments on that issue and in the new design doc.

## Why

Multi-hop chaining silently fabricates rates the user never declared, accumulating uncertainty across vendor / time / spread differences invisibly. Returning `None` and surfacing missing `P` directives is healthier feedback. The TS port for cross-language consumers (#151) drops from ~70 LOC to ~10 LOC as a side effect, which makes the format-as-API claim near-trivially demonstrable.

Full reasoning in [`docs/exchange-rates.md`](https://github.com/alevy/doppio/blob/feat/158-direct-only-fx/docs/exchange-rates.md), including a TypeScript reference implementation and a comparison table with hledger/Beancount/ledger-cli.

## Changes

- `crates/doppio/src/elaboration_ext.rs`:
  - `Journal::exchange_rate_at` rewritten as a linear scan; ~20 LOC of substance
  - Doc comment rewritten to spell out direct + inverse semantics and link the design doc
  - Tests: dropped `_two_hop_chain`; renamed `_no_path_returns_none` → `_no_quote_returns_none`; added `_chained_path_not_traversed` as the explicit "no transitivity" contract
  - Dropped now-unused `HashSet`/`VecDeque` imports
- `proto/doppio.proto`: added "Exchange rate lookup" section to the top comment block with normative algorithm pseudocode and TypeScript / Python / Go reference recipes (mirrors the existing multilingual decimal-reassembly pattern)
- `docs/exchange-rates.md` (new): durable design-decision capture for future high-level documentation. Algorithm, rationale (3 reasons), comparison table with the other PTA tools, TS reference impl, CLI behaviour, history note linking to the [doppio-research embedding-strategies exploration](https://github.com/alevy/doppio-research/blob/main/prototypes/exchange-rates-baking/FINDINGS.md).
- `docs/SUPPORTED_FEATURES.md`: updated `--exchange` row to flag the no-chaining contract and link the design doc

## CLI behaviour

No code changes. PR #155's CLI already handles `None` returns gracefully (stderr warning + leave the commodity unconverted in the report). The practical difference is that more pairs now return `None` legitimately — exactly the chains we're refusing to fabricate.

## Test plan

- [x] `cargo fmt --all --check` — clean
- [x] `cargo test --workspace` — 230 lib + parity + CLI integration tests pass
- [x] `cargo bench --no-run --workspace` — clean
- [x] `cargo build --target wasm32-unknown-unknown -p doppio --lib` — clean
- [ ] CI green

### DIFF
--- a/crates/doppio/src/elaboration_ext.rs
+++ b/crates/doppio/src/elaboration_ext.rs
@@ -212,30 +212,24 @@ impl elaboration::Journal {
             return Some(Decimal::ONE);
         }
 
-        // For each price entry that relates the pair (in either direction), is
-        // within `as_of`, and carries a non-zero price, yield (date, rate).
-        // Take the latest by date. Ties break in favour of the entry that
-        // appears later in source order — `max_by_key` returns the last equal
-        // element, matching "more-recently-declared wins".
         self.prices
             .iter()
+            // Pair must relate, in either direction.
             .filter_map(|hp| {
                 let direct = hp.commodity == from_commodity && hp.price_commodity == to_commodity;
                 let inverse = hp.commodity == to_commodity && hp.price_commodity == from_commodity;
-                if !direct && !inverse {
-                    return None;
-                }
-                let date = hp.date_naive();
-                if as_of.is_some_and(|cutoff| date > cutoff) {
-                    return None;
-                }
-                let raw = hp.price.as_ref()?.to_decimal();
-                if raw.is_zero() {
-                    return None;
-                }
-                let rate = if direct { raw } else { Decimal::ONE / raw };
-                Some((date, rate))
+                (direct || inverse).then(|| (hp, direct, hp.date_naive()))
             })
+            // Date must be at or before as_of (or as_of is None).
+            .filter(|(_, _, date)| as_of.is_none_or(|cutoff| *date <= cutoff))
+            // Price field must be present.
+            .filter_map(|(hp, direct, date)| Some((date, direct, hp.price.as_ref()?.to_decimal())))
+            // Price must be non-zero (avoids degenerate inversions).
+            .filter(|(_, _, raw)| !raw.is_zero())
+            // Apply inversion when the entry was matched in reverse direction.
+            .map(|(date, direct, raw)| (date, if direct { raw } else { Decimal::ONE / raw }))
+            // Latest by date wins. Ties resolve in source order — `max_by_key`
+            // returns the last equal element, matching "more-recently-declared".
             .max_by_key(|(date, _)| *date)
             .map(|(_, rate)| rate)
     }
--- a/crates/doppio/src/elaboration_ext.rs
+++ b/crates/doppio/src/elaboration_ext.rs
@@ -212,30 +212,32 @@ impl elaboration::Journal {
             return Some(Decimal::ONE);
         }
 
-        let mut best: Option<(chrono::NaiveDate, Decimal)> = None;
-        for hp in &self.prices {
-            let direct = hp.commodity == from_commodity && hp.price_commodity == to_commodity;
-            let inverse = hp.commodity == to_commodity && hp.price_commodity == from_commodity;
-            if !direct && !inverse {
-                continue;
-            }
-            let date = hp.date_naive();
-            if as_of.is_some_and(|cutoff| date > cutoff) {
-                continue;
-            }
-            let raw = match hp.price.as_ref() {
-                Some(p) => p.to_decimal(),
-                None => continue,
-            };
-            if raw.is_zero() {
-                continue;
-            }
-            let rate = if direct { raw } else { Decimal::ONE / raw };
-            if best.is_none_or(|(prev_date, _)| date > prev_date) {
-                best = Some((date, rate));
-            }
-        }
-        best.map(|(_, r)| r)
+        // For each price entry that relates the pair (in either direction), is
+        // within `as_of`, and carries a non-zero price, yield (date, rate).
+        // Take the latest by date. Ties break in favour of the entry that
+        // appears later in source order — `max_by_key` returns the last equal
+        // element, matching "more-recently-declared wins".
+        self.prices
+            .iter()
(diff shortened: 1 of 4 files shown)


crates/doppio/src/elaboration_ext.rs
```diff
@@ -215,72 +212,30 @@ impl elaboration::Journal {
...
-        let eligible_quotes = self.prices.iter().filter_map(|hp| {
-            let price_val = hp.price.as_ref()?.to_decimal();
-            if price_val == Decimal::ZERO {
-                return None;
+        let mut best: Option<(chrono::NaiveDate, Decimal)> = None;
+        for hp in &self.prices {
+            let direct = hp.commodity == from_commodity && hp.price_commodity == to_commodity;
+            let inverse = hp.commodity == to_commodity && hp.price_commodity == from_commodity;
+            if !direct && !inverse {
+                continue;
```
alevy: This whole imperitive loop would, again, be _much_ clearer if using iterator filters instead. Logically, this should just be: sort by date (already sorted); filter to only include valid entries where the `commodity` and `price_commodity` are the to and from or from and to commodities; filter by dates that are before as_of if it is present; `next()`.

alevy: Done in 45e79e1 — refactored the imperative best-of-loop into a `filter_map + max_by_key` chain. One subtle behavioural change worth noting: `max_by_key` returns the LAST equal element, so on a same-date tie the entry that appears later in source order now wins ("more-recently-declared wins"), where the imperative version had first-wins. I think the new tie behaviour better matches user intent — if you write the same pair twice on the same date, the later declaration is the override — but happy to flip it back to first-wins if you'd rather; `min_by_key` with reversed key or an early-out version would do it.

alevy: Done in c493cfe — broke the single `filter_map` into one filter / filter_map per invariant. Each step does exactly one job and the pipeline reads top-to-bottom as a checklist:

```
.filter_map  → pair must relate (direct or inverse)
.filter      → date ≤ as_of
.filter_map  → price field must be present
.filter      → price must be non-zero
.map         → invert when matched in reverse direction
.max_by_key  → latest date wins
.map         → unwrap (date, rate) → rate
```

-6 net lines vs the previous closure-soup. Iterator fusion means the runtime cost is unchanged.

Issue #185, opened 2026-05-06
[Beancount] Complete lot annotation handling


alevy:
## Background

#146 lands a best-effort lot annotation parser. The current implementation in \`crates/doppio/src/grammars/beancount/mod.rs::merge_lot_annotation_into\` covers the common case (\`{cost USD, 2024-02-15, \"label\"}\`) but has known gaps that are tracked here rather than blocking #146.

## What's missing

1. **Wildcard \`{*}\` and \`{*, ...}\`**: Beancount's \"automatic cost\" wildcard is silently dropped today. There is no AST representation for \"infer this lot's cost from the inventory.\"

2. **Total-cost form \`{{total}}\`**: The grammar matches \`{{...}}\` separately from \`{...}\` (per the spec, the double-brace form means total-cost rather than per-unit-cost). The current adapter does NOT distinguish between the two -- both feed through \`merge_lot_annotation_into\` identically and end up in \`LotAnnotation::cost\`. For per-unit semantics this is correct; for total-cost it silently misrepresents the lot's basis. The hledger frontend rejects the analogous \`{{total}}\` form at parse time with a clear error message; Beancount probably wants the same fallback until total-cost basis is properly modelled.

3. **Cost-spec arithmetic edge cases**: Beancount allows expressions inside lot specs, currencies as standalone tokens (\"convert at the prevailing rate\"), and various other forms surfaced by real corpora. The current per-comma split + value_expr re-parse won't cover all of them.

4. **Per-currency basis**: Beancount distinguishes the cost commodity from the held commodity (\`10 AAPL {150 USD}\`). Today we only capture the value expression; downstream consumers can't tell which commodity is the cost basis vs the held lot.

## Suggested approach

Most of these are captured by the carve-out at the top of \`beancount.pest\`:
> Beancount's lot syntax inside \`{...}\` is captured as raw text; the AST adapter (#146) is responsible for parsing the inner fields (cost, date, label).

Once we have a Beancount corpus to test against (#183-related: cross-frontend canonical-tool CI), the gaps that show up first should drive the priority order.

## Why this isn't blocking M#9

The Beancount frontend is shipping experimental. \`pad\` (#147) is a bigger semantic gap than lot-annotation completeness; getting that landed and validated comes first. Real downstream usage will surface which lot edge cases actually matter.

alevy:
Closing with a retrospective. Each of the four sub-gaps in the original body is either resolved or explicitly punted; the validation-layer follow-up is split out under #227.

[...]

PR #195, opened 2026-05-06
feat: cross-frontend total-cost lot basis (#193)

### LINKED ISSUE #193: Cross-frontend: support `{{total}}` total-cost lot basis (hledger + Beancount)
## Background

Both ledger-cli/hledger and Beancount support a total-cost lot annotation form alongside the per-unit form:

| Form | Meaning | Example |
|------|---------|---------|
| \`{cost}\` | per-unit cost | \`10 AAPL {\$150}\` -- each share acquired at \$150 |
| \`{{total}}\` | total cost | \`10 AAPL {{\$1500}}\` -- the entire 10-share lot acquired for \$1500 (per-unit \$150) |

The two forms encode the same basis differently. Users reach for \`{{total}}\` when they have a transaction recorded as \"dollars spent on a lot\" rather than \"price per share.\"

## Today (gap)

Neither frontend handles total-cost basis correctly:

- **hledger** rejects \`{{total}}\` at parse time with a clear error (\"double-brace \`{{total}}\` lot syntax is not yet implemented; use \`{cost}\` for per-unit cost or \`@@ total\` for transient total cost instead\"). Hostile-but-loud.
- **Beancount** rejects \`{{total}}\` at parse time as of #194 with the same kind of error (was previously silent miscomputation; the stopgap stops the silent failure mode while this issue is open).

Filed in response to discovering during the Beancount frontend work that we had silently treated hledger as \"complete\" while it was demonstrably incomplete on this surface -- the parser rejection masked the underlying semantic gap.

## Scope

Single PR covers both frontends. Smaller than initially scoped because no proto change is needed: \`{{total}}\` vs \`{cost}\` is a parse-time fact, and the elaborator can always divide by the unit count to produce a canonical per-unit basis that matches the existing wire format.

- **AST** (\`LotAnnotation\`): add \`cost_is_total: bool\` (parse-time-only; defaults to \`false\`).
- **Adapters**: hledger and Beancount both detect the \`{{...}}\` brace count and set the new field. Today's parse-time rejections in both frontends are removed.
- **Elaborator**: when applying a lot annotation with \`cost_is_total=true\`, divide the cost expression by the posting's unit count before using it as per-unit basis. Everything downstream (balance computation, the elaborated wire format) sees only the per-unit form.
- **Tests**: per-unit and total-cost lots produce the same balance for the same posting; the existing per-unit-cost tests still pass; round-trip parity through \`bean-check\` / \`ledger-cli\`.
- **Proto**: no change. The wire format already carries the elaborated per-unit basis.

## Why now

Modest scope (~80-100 lines, single PR). Correct semantics in both frontends in one move. Fixes:

1. Beancount's parse-time rejection (introduced in #194 as a stopgap)
2. hledger's existing parse-time rejection
3. The Beancount \`{{total}}\` carve-out from #185

When this lands, the immediate-stopgap rejections in both frontends become unnecessary and are removed.

## Related

- #185 (Beancount lot annotation completeness umbrella) -- this closes the \`{{total}}\` sub-gap.
- #194 -- the Beancount parse-time rejection stopgap that this issue makes redundant.
- The hledger frontend's \`parse_lot_annotation_into\` rejects \`{{total}}\` today; that rejection is also removed in this PR.

alevy:
## Summary

Closes the `{{total}}` lot annotation gap in both the hledger and Beancount frontends. Until this PR, the form was either rejected at parse time (hledger; Beancount as of #194 stopgap) or silently miscomputed (earlier Beancount). Now both frontends accept it and the elaborator derives the canonical per-unit basis by dividing the captured total by the posting's unit count.

Closes #193 and the `{{total}}` sub-gap of #185.

## Why no proto change

`{{total}}` vs `{cost}` is a parse-time fact about how the source expressed the basis. Once we know the total and the unit count, the elaborator computes the same per-unit basis either way. The proto wire format already carries the per-unit form; it never needed a flag.

This is smaller scope than the original #193 issue body suggested (~100 lines instead of ~150) — that estimate has been corrected on the issue.

## Changes

- **AST** (`LotAnnotation`): new `cost_is_total: bool` (parse-time fact, defaults to false). Internal-only; no public API change.
- **Adapters**: hledger and Beancount both detect the `{{...}}` brace count and set the bool. Both parse-time rejections are removed.
- **Elaborator**: when `cost_is_total` is true, divide the cost by the absolute unit count before storing the per-unit form in the proto Lot. New `ElaborationError::TotalCostWithZeroUnits` for the malformed `{{total}}` + 0-unit posting case.

## Tests

- **Beancount** (`double_brace_total_cost_lot_elaborates_to_per_unit`): a `{{total}}` posting and an equivalent `{cost}` posting produce **identical elaborated lots** end-to-end. Replaces the previous rejection test.
- **hledger** (`test_lot_annotation_double_brace_total_form`): the AST shape correctly captures `cost_is_total = true` for `{{total}}`. Replaces the previous rejection test.
- All existing per-unit-cost tests continue to pass — confirms the per-unit code path is undisturbed.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] `cargo test --workspace` (278 doppio lib + ...; all green)
- [x] `cargo build --target wasm32-unknown-unknown -p doppio --lib`
- [x] `bean-check tests/fixtures/sample.beancount` (still clean)

### DIFF
--- a/crates/doppio/src/elaborator.rs
+++ b/crates/doppio/src/elaborator.rs
@@ -154,6 +154,10 @@ pub enum ElaborationError {
         /// A rendered form of the failing expression (for diagnostics).
         expression: String,
     },
+    /// A `{{total}}` lot annotation was attached to a posting whose
+    /// unit count is zero, so per-unit basis cannot be derived. This is
+    /// a malformed input, not a doppio bug.
+    TotalCostWithZeroUnits,
 }
 
 /// Error produced when evaluating a value expression (e.g., an amount or
@@ -331,6 +335,12 @@ impl Display for ElaborationError {
                 write!(
                     f,
                     "tag assertion failed for {tag_name}: \"{tag_value}\": assert {expression}"
+                )
+            }
+            ElaborationError::TotalCostWithZeroUnits => {
+                write!(
+                    f,
+                    "`{{{{total}}}}` lot cost cannot be applied to a posting with zero units"
                 )
             }
         }
@@ -565,14 +575,28 @@ impl TryFrom<resolution::HIR> for crate::elaboration::Journal {
                                                 .expect("epoch is valid");
                                             let proto_date =
                                                 ann.date.map(|d| (d - epoch).num_days() as i32);
+                                            let cost_is_total = ann.cost_is_total;
                                             let (proto_cost, cost_pair) =
                                                 if let Some(cost_expr) = ann.cost {
-                                                    let (cv, cc) =
+                                                    let (mut cv, cc) =
                                                         evaluator::eval_and_normalize_amount(
                                                             cost_expr,
                                                             entry_context,
                                                             &state,
                                                         )?;
+                                                    // `{{total}}` form: convert total -> per-unit
+                                                    // by dividing by the posting's absolute unit
+                                                    // count. The proto stores the canonical
+                                                    // per-unit form; the source-syntax
+                                                    // distinction is intentionally erased here.
+                                                    if cost_is_total {
+                                                        if value.is_zero() {
+                                                            return Err(
+                                                        ElaborationError::TotalCostWithZeroUnits,
+                                                    );
+                                                        }
+                                                        cv /= value.abs();
+                                                    }
                                                     let proto_amount = crate::elaboration::Amount {
                                                         by_commodity: BTreeMap::from([(
                                                             cc.clone(),
--- a/crates/doppio/src/elaborator.rs
+++ b/crates/doppio/src/elaborator.rs
@@ -155,8 +155,8 @@ pub enum ElaborationError {
         expression: String,
     },
     /// A `{{total}}` lot annotation was attached to a posting whose
-    /// unit count is zero, so per-unit basis cannot be derived. This is
-    /// a malformed input, not a doppio bug.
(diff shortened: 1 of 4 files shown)


crates/doppio/src/elaborator.rs
```diff
@@ -154,6 +154,10 @@ pub enum ElaborationError {
         /// A rendered form of the failing expression (for diagnostics).
         expression: String,
     },
+    /// A `{{total}}` lot annotation was attached to a posting whose
+    /// unit count is zero, so per-unit basis cannot be derived. This is
+    /// a malformed input, not a doppio bug.
```
alevy: Who is the sentence intended for? Why would we state in public documentation that something in our code isn't a bug?

alevy: Good catch — fixed in 826b61d. Removed the second sentence; the first one (describing when the error fires) is enough. Also added a directive to my local rust-feature-writer agent and a feedback memory so I don't write defensive/self-justifying language in public docs again ("this is not a bug", "this is intentional", project-name self-references, etc.).

PR #204, opened 2026-05-07
hledger: support `=*` / `==*` strict-zero balance assignment (#200)

### LINKED ISSUE #200: [hledger] Support `==* 0` strict-zero balance assignment
## Background

Surfaced while expanding the parity corpus (#197): hledger's \`examples/multicurrency.journal\` uses

\`\`\`hledger
2015-12-31 retain earnings
    Income                                               ==* 0
    Equity:Retained Earnings
\`\`\`

The \`==*\` operator is hledger's \"strict zero\" balance assignment: assert that the account's balance is zero in EVERY commodity it has held (not just one), and synthesize whatever postings are needed to make that true. Used in fiscal-year-end retained-earnings workflows.

Our hledger grammar currently parses \`==\` (strict same-commodity assertion) but rejects \`==*\`:

\`\`\`
ParsingError { line: \"    Income                                               ==* 0\" }
\`\`\`

## Scope

- Grammar: extend the \`assertion_op\` / balance-assignment rule to recognise \`==*\` alongside \`==\` and \`=\`.
- Adapter: capture the \"all-commodity\" flag on the \`AmountDetails\`.
- Elaborator: implement the multi-commodity zero-out semantics (synthesize one balancing posting per commodity in the account).

The grammar change is small. The elaborator change is larger -- it needs to know the account's full commodity inventory at the assertion's point in time, which the running state tracks but isn't currently consumed for inference.

## Why post-#197

Discovered while pulling \`multicurrency.journal\` into the corpus. The grammar gap alone is small; the elaborator semantics are real work. Filing as a follow-up keeps the corpus expansion focused and lets this issue scope the elaborator change properly.

## Related

- #197 (corpus expansion)
- The strict (\`==\`) vs weak (\`=\`) distinction is already handled at the AST level; this just adds a third variant.

alevy:
## Summary

The strict-zero across all commodities form, used in fiscal-year retained-earnings transactions:

\`\`\`hledger
2024-12-31 retain earnings
    Income:Salary    ==* 0
    Equity:Retained-Earnings
\`\`\`

Previously the grammar only recognised \`=\` and \`==\`; an \`==* 0\` line parse-errored. With this PR both \`=*\` and \`==*\` are accepted, and the elaborator synthesizes a multi-commodity posting that brings every currently-held commodity in the account to the target value.

Closes #200.

## Changes

- **Grammar**: \`assertion_op\` extended to \`(\"==\" | \"=\") ~ \"*\"?\` so the optional all-commodities qualifier is captured.
- **AST**: new \`AmountDetails::BalanceAssignmentAllCommodities(ValueExpr)\` variant. Display impl renders as \`==* X\`.
- **Adapter**: detects the trailing \`*\` on the op, routes to the new variant.
- **Elaborator**: when processing the new variant, snapshots the account's running per-commodity inventory, computes per-commodity \`delta = target - current_balance\`, and synthesizes ONE multi-commodity Posting that brings each currently-held commodity to the target. Updates \`transaction_state\` and \`account_balances\` per commodity so null-posting absorption + balance accounting downstream see the synthesized values correctly. Proto schema unchanged (multi-commodity Amounts already supported).

## Tests

- \`assertion_op_recognises_star_qualifier\` -- both \`==*\` and \`=*\` parse to the new \`BalanceAssignmentAllCommodities\` variant
- \`tests/parity/hledger-zerostar.journal\` -- new parity fixture (USD + EUR salaries zeroed into Equity:Retained-Earnings); hledger and doppio agree end-to-end

280 doppio lib tests, +2 new.

## What this does NOT do

hledger upstream's \`multicurrency.journal\` also uses \`comment ... end comment\` block-comment syntax which is a separate, unrelated gap. This PR unblocks the \`==*\` axis only; the comment-block gap stays open.

## Test plan

- [x] \`cargo fmt --all -- --check\`
- [x] \`cargo test --workspace\` (280 doppio lib + 49 elaborator + ...; all green)
- [x] Parity (5 positive incl. new \`hledger:zerostar\`, 3 negative; all green locally)

### DIFF
--- a/crates/doppio/src/ast.rs
+++ b/crates/doppio/src/ast.rs
@@ -575,8 +575,12 @@ impl Display for AmountDetails {
             AmountDetails::BalanceAssignment(value) => {
                 write!(f, "={value}")
             }
+            // The all-commodities qualifier is hledger-specific syntax;
+            // the `Display` impl is otherwise ledger-syntax-shaped (used
+            // by `write_ledger`), so render lossy as a single-commodity
+            // assignment. Per-frontend output is a separate concern.
             AmountDetails::BalanceAssignmentAllCommodities(value) => {
-                write!(f, "==* {value}")
+                write!(f, "={value}")
             }
         }
     }
--- a/crates/doppio/src/ast.rs
+++ b/crates/doppio/src/ast.rs
@@ -575,12 +575,15 @@ impl Display for AmountDetails {
             AmountDetails::BalanceAssignment(value) => {
                 write!(f, "={value}")
             }
-            // The all-commodities qualifier is hledger-specific syntax;
-            // the `Display` impl is otherwise ledger-syntax-shaped (used
-            // by `write_ledger`), so render lossy as a single-commodity
-            // assignment. Per-frontend output is a separate concern.
+            // hledger-specific syntax with no ledger-cli equivalent:
+            // ledger's `= X` is single-commodity assignment, which is
+            // semantically different from "zero every commodity in
+            // the inventory." Render as `==* X` (the original
+            // hledger source form); output containing this variant
+            // therefore does NOT round-trip through the ledger
+            // parser. Proper per-frontend writers are #206.
             AmountDetails::BalanceAssignmentAllCommodities(value) => {
-                write!(f, "={value}")
+                write!(f, "==* {value}")
             }
         }
     }
--- a/crates/doppio/src/ast.rs
+++ b/crates/doppio/src/ast.rs
@@ -521,6 +521,12 @@ pub enum AmountDetails {
     /// A balance assignment: `= target_balance`, with no explicit debit/credit
     /// value. The posting amount is computed as `target - current_balance`.
     BalanceAssignment(ValueExpr),
+    /// hledger's "balance everything" form: `=* target` or `==* target`,
+    /// typically `==* 0` in retained-earnings transactions. Synthesizes a
+    /// multi-commodity posting that brings the account's balance in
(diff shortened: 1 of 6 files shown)


alevy:
Both review concerns addressed in 1986f2e:

1. `Display` impl no longer emits hledger-specific `==*` syntax. The all-commodities variant now renders lossy as `={value}` (matching `BalanceAssignment`), with a comment noting that the qualifier is dropped at this layer. Per-frontend output is the right long-term fix and is now tracked in **#206** (cross-frontend writers for hledger and Beancount).

2. The `Some("__all_commodities__")` sentinel is gone. New helper `evaluator::eval_amount_value` evaluates a value expression and returns just the `Decimal`, no commodity-normalisation step. The `==*` callsite uses it directly -- no fake commodity name needed.

alevy:
Walking back the first half of my previous fix on your prompting:

`=X` in ledger is single-commodity assignment, which is semantically *different* from `==* X` (zero every commodity). My "lossy" rendering was actually *wrong*. Neither ledger nor Beancount has an all-commodities form -- the only faithful translation is to expand to N explicit postings, which a stateless `Display` impl doesn't have inventory access for.

Reverted to `==* X` in dd93d27 with a doc note that the variant does not round-trip through the ledger parser. The Display-emits-ledger-syntax framing is itself the leaky default; #206 will close it properly with per-frontend writers.

Concern 2 (the `__all_commodities__` sentinel) was correctly addressed in 1986f2e via the new `evaluator::eval_amount_value` helper -- that one stays.

Issue #144, opened 2026-04-29
[Beancount] Frontend implementation (parent)


alevy:
## Goal

Add Beancount as a third experimental frontend, alongside ledger-cli
(canonical) and hledger (experimental v0.3+). Recognised by the
\`.beancount\` file extension; selected automatically by \`dop\` and by
\`doppio::frontend_for_extension\`.

The \`Frontend\` trait shipped in Phase B (#102) makes the wiring
mechanical — but the \`pad\` directive needs new evaluator logic with
no clean ledger-cli analogue (per \`docs/requirements.md\` §12 and the
research feasibility analysis at
\`alevy/doppio-research/frontends-report.md\`).

## Background

Beancount is similar in spirit to hledger but with stricter typing and
a few directives that don't map onto the existing AST:

- \`open\` / \`close\` (account lifecycle)
- \`commodity\` (nearly identical to ledger-cli)
- \`txn\` / no-keyword transactions
- \`pad\` (insert balancing posting up to a balance assertion — the
  novel one)
- \`balance\` (assertion at a specific date)
- \`price\` (P-style)
- \`note\`, \`document\`, \`event\`, \`query\`, \`custom\`

The frontends-report concluded most of the surface fits the existing
AST without changes; \`pad\` is the standout.

## Sub-tasks

- #145 — Beancount pest grammar
- #146 — Beancount AST adapter (Frontend trait impl)
- #147 — \`pad\` directive evaluator design + implementation

## Acceptance

- \`.beancount\` files compile via \`dop\` and through the library API
- A 12-entry sample journal in \`crates/doppio/tests/fixtures/sample.beancount\`
  demonstrating each supported directive
- \`pad\` directive works for the canonical case (open balance pad-in)
- \`docs/SUPPORTED_FEATURES.md\` gains a Beancount section listing
  supported directives and known limitations
- ~~Marked **experimental** in the docs — not held to the same
  semver-stable surface as the ledger frontend~~
  *(Struck out 2026-05-07 — Beancount frontend has converged with
  ledger-cli and hledger in coverage, pad/balance correctness, and
  parity-test pass rate; no longer materially more experimental than
  the other two frontends.)*

PR #225, opened 2026-05-07
beancount: pad covers every asserted commodity on the same target (#220)

### LINKED ISSUE #220: [beancount] `pad` only synthesises one commodity; multi-commodity assertions on the same target fail
## Background

Surfaced while expanding the parity corpus (#197). Vendoring \`beancount/beancount@5704a86:examples/simple/basic.beancount\` triggered a balance-assertion failure that bean-check 3.2.0 accepts cleanly.

## Minimal repro

\`\`\`beancount
option \"operating_currency\" \"USD\"

1970-01-01 open Assets:Cash
1970-01-01 open Equity:Opening-Balances

2007-12-30 pad  Assets:Cash  Equity:Opening-Balances

2007-12-31 balance  Assets:Cash    200 CAD
2007-12-31 balance  Assets:Cash    300 USD
\`\`\`

\`bean-check\`: exit 0. Per-account positions show Assets:Cash holding both 200 CAD and 300 USD; Equity:Opening-Balances offsets each.

\`dop balance\`:

\`\`\`
Error: BalanceAssertionFailed { account: \"Assets:Cash\", date: 2007-12-31,
expected_amount: 300, expected_commodity: \"USD\", actual_amount: 0 }
\`\`\`

## Diagnosis

\`pad\` currently fires once: it consumes itself for the *first* balance assertion that follows on the target account. Subsequent balance assertions on the same target (in a different commodity) see no pending pad and fail because the running balance is zero.

Beancount's \`pad\` is per-commodity: a single \`pad\` covers every balance assertion on the same account up to the next non-zero direct posting in any commodity.

## Where to look

- \`crates/doppio/src/elaborator.rs:420\` -- \`state.pending_pads.remove(&assertion.account)\` consumes the pad on first assertion. The pad needs to be conditionally retained when there's more state to cover, or pre-expanded at pad-evaluation time once we know what assertions are upcoming.
- \`crates/doppio/src/grammars/beancount/mod.rs\` -- existing pad tests cover single-commodity only. Add a multi-commodity probe.

## Tests

- Multi-commodity pad with two same-day assertions on the same target.
- Three-commodity case.
- Pad that's *only* needed for one commodity (other commodities have direct postings reaching the asserted amount) -- pad should still synthesize for the unfunded commodity.

## Why this surfaced from #197

\`basic.beancount\` is a 643-line realistic example with pad+balance for Cash in both CAD and USD. Real-world Beancount files routinely use \`pad\` against multi-currency accounts; the gap is significant.

## Vendoring blocker

Once fixed, vendor \`basic.beancount\` as \`tests/parity/beancount-basic.beancount\` and add to \`POSITIVE\` -- the rest of its features (lots, prices, pushtag/poptag, event directive) parse cleanly.

## Related

- #144 -- Beancount frontend umbrella.
- #197 -- corpus expansion (this is one of the gaps it surfaces).

alevy:
## Summary

A single \`pad Account Source\` followed by multiple balance assertions on the same target in different commodities now synthesises one pad transaction per asserted commodity, matching bean-check 3.2.0. Previously, the pad was consumed on the first assertion; later assertions in other commodities saw running balance = 0 and erroneously failed.

## The change

One line in \`crates/doppio/src/elaborator.rs\`: \`state.pending_pads.remove(&account)\` → \`state.pending_pads.get(&account).cloned()\`. The pad stays pending; each assertion in a different commodity computes its own delta from a zero baseline and synthesises its own pad transaction. Repeated assertions in an already-padded commodity see \`actual == expected\`, compute \`diff = 0\`, and the existing \`if !diff.is_zero()\` guard skips synthesis. No double-pad.

The \"most recent pad wins\" semantic still works because a new pad on the same target overwrites the entry in \`pending_pads\`. End-of-journal cleanup (drop unused pads silently) is unchanged.

## Tests

- \`pad_covers_every_asserted_commodity_on_same_target\` -- pad + same-day CAD + USD assertions; verifies two pad transactions, each placing its delta on the literal pad target in the right commodity.
- \`pad_does_not_double_synthesise_for_repeated_commodity\` -- same pad, two same-commodity assertions; verifies exactly one pad transaction.
- New parity fixture \`tests/parity/beancount-basic.beancount\` vendored from \`beancount/beancount@5704a86:examples/simple/basic.beancount\`. Real 643-line Beancount file using multi-currency pad on Assets:Cash plus lots, prices, pushtag/poptag, events.
- All existing pad tests (\`pad_fills_gap_to_next_balance_assertion\`, \`pad_without_following_balance_is_silently_dropped\`, \`most_recent_pad_wins_when_multiple_precede_one_balance\`, \`pad_with_zero_gap_emits_no_transaction\`) still pass.

## Test plan

- [x] \`cargo fmt --check\`
- [x] \`cargo test --workspace\` -- 463 tests pass (was 461; +2 new tests).
- [x] \`python3 scripts/parity_check.py\` -- all 11 fixtures pass against bean-check 3.2.0, hledger 1.40, ledger 3.3.2 (this branch is from main, so doesn't include the 3 fixtures in PR #221; those will rebase on merge).

## Related

- #197 -- corpus expansion (the gap was surfaced by attempting to vendor \`basic.beancount\`).
- #144 -- Beancount frontend umbrella.

Closes #220.

### DIFF
--- a/crates/doppio/src/elaborator.rs
+++ b/crates/doppio/src/elaborator.rs
@@ -60,12 +60,19 @@ struct RunningState {
 }
 
 /// A pad directive observed but not yet consumed by a balance assertion.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 struct PendingPad {
     /// The pad's own date -- used as the synthesized transaction's date.
     date: chrono::NaiveDate,
     /// The counter-account that will absorb the padding amount.
     source_account: String,
+    /// Commodities for which this pad has already fired a synthesised
+    /// transaction. Beancount fires the pad once per commodity at the
+    /// first balance assertion that references it; subsequent
+    /// assertions in the same commodity check the running balance
+    /// against the asserted amount without re-padding. Different
+    /// commodities can each consume the pad once.
+    padded_commodities: std::collections::BTreeSet<String>,
 }
 
 /// A commodity name (e.g. `"USD"`, `"BTC"`, `"$"`).
@@ -382,15 +389,18 @@ impl TryFrom<resolution::HIR> for crate::elaboration::Journal {
             let entry_context = &value.contexts[entry.context_id];
             match entry.data {
                 resolution::Entry::Pad(p) => {
-                    // Beancount `pad`: remember the (target, source) pair until
-                    // the next balance assertion on `target` consumes it. A
-                    // second pad on the same target before any assertion
-                    // overwrites the first (most-recent-pad-wins).
+                    // Beancount `pad`: remember the (target, source) pair so
+                    // upcoming balance assertions on `target` can each fire
+                    // it once per commodity. A second pad on the same target
+                    // overwrites the first (most-recent-pad-wins) -- the new
+                    // pad's `padded_commodities` set starts empty, re-arming
+                    // it for every commodity.
                     state.pending_pads.insert(
                         p.target_account,
                         PendingPad {
                             date: p.date,
                             source_account: p.source_account,
+                            padded_commodities: std::collections::BTreeSet::new(),
                         },
                     );
                 }
@@ -426,7 +436,8 @@ impl TryFrom<resolution::HIR> for crate::elaboration::Journal {
                         }
                     };
 
-                    // If a pad on this account is pending, synthesize a
+                    // If a pad on this account is pending and has not yet
+                    // fired for the asserted commodity, synthesize a
                     // balancing transaction (back-dated to the pad's date)
                     // that brings `actual_amount` up to `expected_amount`,
                     // and apply it to the running state so the assertion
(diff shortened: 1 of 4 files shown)


crates/doppio/src/elaborator.rs
```diff
@@ -433,7 +433,18 @@ impl TryFrom<resolution::HIR> for crate::elaboration::Journal {
                     // below passes by construction. The synthesized
                     // transaction is also pushed into the output journal so
                     // downstream consumers see the inserted postings.
-                    if let Some(pad) = state.pending_pads.remove(&assertion.account) {
```
alevy, line 436: OK, but when _do_ the pads get removed? Shouldn't they be removed at _some_ point?

alevy:
Good catch. You were right -- the previous patch was incomplete: it never consumed pads at all, so a sequence like

\`\`\`beancount
pad Assets:Cash Equity:Open
balance Assets:Cash 100 USD
2024-02-01 * "Spend"
    Assets:Cash    -50 USD
    Expenses:Food
balance Assets:Cash 25 USD     ; bean-check FAILS this; doppio used to silently pass
\`\`\`

silently succeeded because the pad re-fired at the second assertion to bring running from 50 down to 25. bean-check rejects.

Pushed a fix in \`3fbbbc9\`. The correct semantic, verified empirically against bean-check 3.2.0 with four directed probes:

> Each pad fires exactly once *per commodity*. Same-commodity assertions after the first don't re-fire. Different-commodity assertions can each fire once. Real postings between pad and balance don't consume the pad in any commodity -- the running-balance diff already accounts for them.

Implementation: \`PendingPad\` now carries a \`padded_commodities: BTreeSet<String>\`. A new pad on the same target overwrites with an empty set (re-arms every commodity); each assertion fires only if its commodity isn't in the set, then records the firing.

Two new regression tests pinned: the exact bug you asked about (\`pad_does_not_re_fire_after_real_posting_in_same_commodity\`) and the per-commodity independence case (\`pad_per_commodity_independent_of_other_commodity_postings\`). All existing pad tests still pass; all 11 parity fixtures still green.

To answer the original question directly: pads are *never* removed from \`pending_pads\` within a parse. They sit there with a per-commodity firing record. New pad on same target overwrites; end-of-journal drops the rest silently. Whether to garbage-collect long-pending pads on first real posting (different but related semantic concern -- "stale pad that was never asserted") is a separate question I haven't probed; happy to do so if you'd like.

Issue #237, opened 2026-05-07
[elaborator] Per-lot inventory in the running state (foundational; enables phantom-lot validation, auto-booking, per-lot reports)


alevy:
## Articulation

All three reference tools (ledger-cli, hledger, Beancount) model an account as an inventory of distinct lots, keyed by \`(commodity, cost, date, label)\`. doppio's elaborator aggregates per-\`(account, commodity)\`. The wire format \`Posting.lot\` already preserves per-posting lot data; the divergence is only in the elaborator's *running state* during pass-through.

The same fix unblocks several distinct gaps:

1. **Phantom-lot validation** — see #227.
2. **Automatic lot booking** (FIFO / LIFO / AVERAGE on \`open Account \"...\"\`) — Beancount feature; useful for any real brokerage workflow. Not niche, just absent from current corpora. Filed separately as a child of this issue.
3. **Accurate per-lot capital-gains computation** — today doppio's hledger frontend (#210) synthesises gains via a transaction-local \`cost_basis_state\` hack; with per-lot inventory the gains would derive from the actually-matched lot, which is what the canonical tools do.
4. **Per-lot reports** — \`bal --lots\` style output. All three reference tools expose this; doppio can't because there's no per-lot data in the running state to query.

## Empirical context

Probed on a phantom-lot journal:

\`\`\`
2024-01-15 * Buy
    Assets:Coins   10 GOLD {1500.00 USD}
    Assets:Cash   -15000.00 USD
2024-02-15 * Buy more
    Assets:Coins   10 GOLD {1600.00 USD}
    Assets:Cash   -16000.00 USD
2024-03-15 * Sell, claiming a $1700 lot I never bought
    Assets:Coins  -10 GOLD {1700.00 USD, 2024-01-15} @ 1750.00 USD
    Assets:Cash   17500.00 USD
    ...
\`\`\`

| Tool | Accepts? |
|------|----------|
| Beancount | ❌ rejects (\"lot not found in inventory\") |
| ledger-cli | ✅ accepts |
| hledger | ✅ accepts (when balanced via @price-only) |
| doppio | ✅ accepts |

So doppio's loose behaviour matches ledger-cli and hledger today; only Beancount enforces lot existence. \"Adhere to what everyone else is doing\" cuts a single way: matching Beancount means becoming stricter than ledger-cli / hledger.

## Why this isn't a per-frontend asymmetry (the data structure)

[...]

PR #240, opened 2026-05-07
refactor: decouple elaboration semantics from frontend syntax (closes #239)

### LINKED ISSUE #239: [refactor] Decouple elaboration semantics from frontend syntax: explicit `ElaborationConfig` parameter to `elaborate()`
## The architectural separation

The compiler pipeline is three stages: **parse** (text → AST) → **resolve** (AST → HIR; canonicalise dates, expand aliases, capture journal-level overrides) → **elaborate** (HIR → Journal; apply transaction-balance semantics, synthesise residuals, validate assertions). Each stage's job is logically independent of the others.

Today, the frontends (parsers) **inject elaboration semantics into the HIR**. \`BeancountFrontend::parse\` does:

\`\`\`rust
hir.global_context.tolerance_mode = FractionOfSmallestPrecision(0.5);
hir.global_context.balance_mode = CostBasis;
hir.global_context.assertion_scope = Subtree;
\`\`\`

This bakes \"file in beancount syntax\" → \"beancount-style elaboration semantics\" as a hardcoded coupling. It conflates the two stages: the parser's job is *only* to produce an AST that round-trips the source text; the elaborator's job is to apply a chosen set of semantic rules. Whether a journal in beancount syntax is elaborated under beancount, hledger, or ledger-cli rules ought to be a *separate* choice.

Concrete reasons why this matters:

- A user could legitimately want to parse beancount syntax but use ledger-cli's strict balance semantics (no per-transaction tolerance) — e.g. for a regression-finding workflow that wants to surface any rounding looseness.
- Mixed-mode experiments: parse hledger syntax, use beancount-style subtree balance assertions.
- Custom: a user wants their own \`gains_account\` (not the default \`Income:Capital Gains\`) on a per-elaboration basis.
- Tests that exercise the elaborator directly (\`crates/doppio/src/grammars/beancount/mod.rs:1880-1900\` and analogues) currently have to mirror each frontend's defaults by hand-editing \`hir.global_context\` after \`parse()\` returns. With the config explicit, they'd just pass the matching const.

## Naming

Originally I called the proposed grouping \`FrontendSemantics\`, which is the wrong name twice over: the semantics aren't a property of the frontend, and the frontend's job has nothing to do with them. The right name reflects what the struct actually is — a configuration of the elaborator's semantic choices.

\`\`\`rust
pub struct ElaborationConfig {
    pub tolerance_mode: ToleranceMode,
    pub balance_mode: BalanceMode,
    pub assertion_scope: AssertionScope,
    // pub lot_validation_mode: LotValidationMode,  // forthcoming, #237
}
\`\`\`

## API shape

\`\`\`rust
// Today:
pub fn elaborate(hir: HIR) -> Result<Journal, ElaborationError>;

// Proposed:
pub fn elaborate(hir: HIR, config: &ElaborationConfig) -> Result<Journal, ElaborationError>;
\`\`\`

\`Frontend::parse\` returns just the HIR — no longer mutates semantic-config fields on \`global_context\`. Callers pick a config explicitly at elaboration time:

\`\`\`rust
let hir = BeancountFrontend.parse(input, base_path, opener)?;
let journal = elaborate(hir, &BEANCOUNT_DEFAULTS)?;

// or, mix-and-match:
let hir = BeancountFrontend.parse(input, base_path, opener)?;
let journal = elaborate(hir, &LEDGER_DEFAULTS)?;
\`\`\`

Three module-level statics, one per canonical tool, give the ergonomic default path:

\`\`\`rust
pub static LEDGER_DEFAULTS: LazyLock<ElaborationConfig> = LazyLock::new(|| ElaborationConfig {
    tolerance_mode: ToleranceMode::Strict,
    balance_mode: BalanceMode::CostBasis,
    assertion_scope: AssertionScope::Direct,
});

pub static HLEDGER_DEFAULTS: LazyLock<ElaborationConfig> = LazyLock::new(|| ElaborationConfig {
    tolerance_mode: ToleranceMode::Strict,
    balance_mode: BalanceMode::AtPriceWithSynthesis {
        gains_account: \"Income:Capital Gains\".into(),
    },
    assertion_scope: AssertionScope::Direct,
});

pub static BEANCOUNT_DEFAULTS: LazyLock<ElaborationConfig> = LazyLock::new(|| ElaborationConfig {
    tolerance_mode: ToleranceMode::FractionOfSmallestPrecision(dec!(0.5)),
    balance_mode: BalanceMode::CostBasis,
    assertion_scope: AssertionScope::Subtree,
});
\`\`\`

(Or const-eval'd if the field types allow; \`gains_account: String\` likely requires \`LazyLock\`. \`Cow<'static, str>\` would let it be a const default that the user can still own. Implementation detail.)

A high-level convenience for the common case:

\`\`\`rust
pub fn compile_with_defaults(source: &str, frontend: impl Frontend) -> Result<Journal, ...> {
    let hir = frontend.parse(source, ..., ...)?;
    let config = match frontend.kind() { /* match -> &*BEANCOUNT_DEFAULTS, etc. */ };
    elaborate(hir, config)
}
\`\`\`

## Journal-injected overrides stay separate

This is the second piece I got wrong originally. \`tolerance_overrides\` (populated by Beancount's \`option \"inferred_tolerance_default\"\` directive) is **not** a frontend-default selector — it's a per-journal override that the resolver builds from directives. It belongs on \`GlobalContext\` (mutable, per-journal HIR state), not in \`ElaborationConfig\` (immutable, per-elaboration-call).

The layering at elaboration time:

\`\`\`rust
fn effective_tolerance(commodity: &str, hir: &HIR, config: &ElaborationConfig) -> ToleranceMode {
    // 1. Per-commodity override from a journal directive wins.
    if let Some(override_value) = hir.global_context.tolerance_overrides.get(commodity) {
        return ToleranceMode::Absolute(*override_value);
    }
    // 2. Otherwise the config's default applies.
    config.tolerance_mode
}
\`\`\`

\`commodity_properties\`, \`account_properties\`, \`tag_properties\` are the same shape: journal-derived state, not config. They stay on \`GlobalContext\`.

## Migration shape

1. Define \`ElaborationConfig\`. Move the four (eventually five) frontend-default fields off \`GlobalContext\` into it.
2. Define the three \`*_DEFAULTS\` statics.
3. Change \`elaborate()\`'s signature to take \`&ElaborationConfig\`.
4. Each \`Frontend::parse\` impl removes its \`hir.global_context.X = ...\` lines.
5. Each call site of \`elaborate()\` passes the appropriate config.
6. The HIR-bypass tests (\`grammars/*/mod.rs\` test helpers) update to pass a config explicitly instead of mirroring fields on \`hir.global_context\`.

## Why land alongside #237

#237 introduces a fifth selector (\`LotValidationMode\`). If we add it before this refactor, the accumulation continues; if after, the new field slots cleanly into \`ElaborationConfig\` from day one. Two-PR sequence:

1. **This issue (#239)** — pure refactor; behaviour identical.
2. **#237** — adds \`lot_validation_mode: LotValidationMode\` to \`ElaborationConfig\`, threads it through, lands the lot-aware running state.

## Acceptance

- \`elaborate()\` takes \`&ElaborationConfig\`. Frontends no longer mutate \`global_context\` to set semantic defaults.
- Three \`*_DEFAULTS\` statics, one per canonical tool.
- All existing tests pass without behaviour change.
- HIR-bypass tests use the matching const directly instead of hand-mirroring fields.
- A new doctest or integration test demonstrates the cross-syntax-semantics use case (parse beancount, elaborate with \`LEDGER_DEFAULTS\`) to make the decoupling concrete.

## Out of scope

- Renaming the existing types (\`ToleranceMode\`, etc.). Cosmetic; can ride along but isn't required.
- Visibility tightening on \`GlobalContext\` fields.
- Per-account / per-tag overrides of elaboration config (would require new directive support; not implied by this refactor).

## Related

- #237 — per-lot inventory; benefits from this landing first
- #210 / #214 / #198 — the original landed selectors that this consolidates

alevy:
Closes #239. Pure refactor — behaviour identical, all 14 parity fixtures still pass.

## What changes

The compiler pipeline is **parse → resolve → elaborate**. Each stage has its own job: parsers tokenise dialects into ASTs, the resolver canonicalises, the elaborator applies a semantic ruleset. Until now, the frontends muddled the layers by injecting per-tool semantic defaults directly onto \`GlobalContext\`:

\`\`\`rust
hir.global_context.tolerance_mode = FractionOfSmallestPrecision(0.5);
hir.global_context.balance_mode = CostBasis;
hir.global_context.assertion_scope = Subtree;
\`\`\`

That baked \"file in beancount syntax\" → \"beancount-style elaboration semantics\" as a hardcoded coupling. The pairing is conventional but **not fundamental**: a user could legitimately want to parse beancount syntax under ledger-cli rules, hledger syntax under beancount rules, or any mix-and-match.

This PR factors out the syntax → semantics pairing as a separate concern.

## Five changes

1. **New \`ElaborationConfig\`** in \`resolution\` — the three per-tool selectors that used to live on \`GlobalContext\` (\`tolerance_mode\`, \`balance_mode\`, \`assertion_scope\`).
2. **\`pub fn elaborate(hir, &ElaborationConfig)\`** is now the single entry point. The old \`TryFrom<HIR> for Journal\` impl is gone — converting an HIR to a Journal isn't a one-arg operation any more, it requires choosing a semantic ruleset.
3. **\`Frontend::defaults() -> &'static ElaborationConfig\`** is a new trait method. Each frontend returns the matching tool's defaults — a *convenience pairing*, not a forced coupling.
4. **\`static *_DEFAULTS: LazyLock<ElaborationConfig>\`** in each frontend module (\`LEDGER_DEFAULTS\`, \`HLEDGER_DEFAULTS\`, \`BEANCOUNT_DEFAULTS\`).
5. **Frontends no longer mutate semantic-config fields**. \`Frontend::parse\` produces just the parse + resolve output.

\`tolerance_overrides\` on \`GlobalContext\` stays put — it's *journal* state populated from \`option\` directives during resolution, not config. At elaboration time, per-commodity overrides layer on top of the config's \`tolerance_mode\`.

## Composability

\`\`\`rust
let hir = BeancountFrontend.parse(input, base_path, opener)?;
let journal = elaborate(hir, BeancountFrontend.defaults())?;     // canonical pairing
let journal = elaborate(hir, LedgerFrontend.defaults())?;        // mix-and-match
let journal = elaborate(hir, &my_custom_config)?;                // arbitrary
\`\`\`

## API impact (breaking)

- \`pub fn elaborate(hir)\` → \`pub fn elaborate(hir, &ElaborationConfig)\`. Caller passes a config explicitly.
- \`compile()\` is type-locked to ledger-cli (takes a \`grammars::ledger::Parser\`); internally uses \`LEDGER_DEFAULTS\`. Caller-facing signature unchanged.
- \`eval_transaction()\` (programmatic builder API) uses \`LEDGER_DEFAULTS\` internally since no frontend is involved. Caller-facing signature unchanged.
- \`TryFrom<HIR> for Journal\` is gone — there's no automatic conversion any more.

The \`elaborate\` change is the only caller-visible breakage.

## Files touched

\`crates/doppio/src/resolution.rs\`, \`elaborator.rs\`, \`frontend.rs\`, \`grammars/{ledger,hledger,beancount}/mod.rs\`, \`lib.rs\`; \`crates/doppio-cli/src/main.rs\`, \`src/bin/timings.rs\`, \`tests/dop_format.rs\`; \`crates/doppio-categorize/examples/eval_holdout.rs\`.

## Test plan

- [x] \`cargo fmt --check\`
- [x] \`cargo test --workspace\` — 304 lib tests pass.
- [x] \`python3 scripts/parity_check.py\` — all 14 fixtures pass against bean-check 3.2.0, hledger 1.40, ledger 3.3.2.
- [x] \`python3 scripts/parity_check.py --self-test\` — 12/12 comparator assertions hold.

## Why this lands before #237

#237 (per-lot inventory) introduces a fifth selector (\`LotValidationMode\`). With this refactor in first, that selector slots cleanly into \`ElaborationConfig\` from day one — and reviewers of #237 can see the lot work as pure new functionality on top of an already-tidy structure rather than refactor + new field in one PR.

Closes #239.

### DIFF
--- a/crates/doppio-categorize/examples/eval_holdout.rs
+++ b/crates/doppio-categorize/examples/eval_holdout.rs
@@ -47,7 +47,7 @@ fn load_journal(path: &Path) -> Result<Journal, Box<dyn std::error::Error>> {
         let mut input = String::new();
         File::open(path)?.read_to_string(&mut input)?;
         let hir = frontend.parse(&input, base_path, &doppio::file_opener)?;
-        Ok(doppio::elaborate(hir, frontend.defaults())?)
+        Ok(doppio::elaborate(hir, &frontend.elaboration_defaults())?)
     }
 }
 
--- a/crates/doppio-categorize/examples/eval_holdout.rs
+++ b/crates/doppio-categorize/examples/eval_holdout.rs
@@ -47,7 +47,7 @@ fn load_journal(path: &Path) -> Result<Journal, Box<dyn std::error::Error>> {
         let mut input = String::new();
         File::open(path)?.read_to_string(&mut input)?;
         let hir = frontend.parse(&input, base_path, &doppio::file_opener)?;
-        Ok(hir.try_into()?)
+        Ok(doppio::elaborate(hir, frontend.defaults())?)
     }
 }
 
--- a/crates/doppio-cli/src/bin/timings.rs
+++ b/crates/doppio-cli/src/bin/timings.rs
@@ -43,7 +43,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // --- elaboration ---
     // Expression evaluation, transaction balancing, account registration.
     let t3 = Instant::now();
-    let journal: doppio::elaboration::Journal = doppio::elaborate(hir)?;
+    // The parser is ledger-cli; use the matching default semantics.
+    let journal: doppio::elaboration::Journal =
+        doppio::elaborate(hir, &doppio::grammars::ledger::LEDGER_DEFAULTS)?;
     eprintln!("elaboration: {:>8.3}s", t3.elapsed().as_secs_f64());
 
     // --- serialize ---
--- a/crates/doppio-cli/src/bin/timings.rs
+++ b/crates/doppio-cli/src/bin/timings.rs
@@ -45,7 +45,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let t3 = Instant::now();
     // The parser is ledger-cli; use the matching default semantics.
     let journal: doppio::elaboration::Journal =
-        doppio::elaborate(hir, &doppio::grammars::ledger::LEDGER_DEFAULTS)?;
+        doppio::elaborate(hir, &doppio::grammars::ledger::ledger_defaults())?;
     eprintln!("elaboration: {:>8.3}s", t3.elapsed().as_secs_f64());
 
     // --- serialize ---
--- a/crates/doppio-cli/src/main.rs
+++ b/crates/doppio-cli/src/main.rs
@@ -232,12 +232,12 @@ fn load_proto_journal_with_tolerance(
         // Start from the matching tool's default semantics; override the
         // tolerance fraction if `--tolerance` was passed on the CLI.
         let config = match tolerance_override {
-            None => frontend.defaults().clone(),
+            None => frontend.elaboration_defaults(),
             Some(fraction) => doppio::resolution::ElaborationConfig {
                 tolerance_mode: doppio::resolution::ToleranceMode::FractionOfSmallestPrecision(
                     fraction,
                 ),
-                ..frontend.defaults().clone()
+                ..frontend.elaboration_defaults()
             },
         };
         Ok(doppio::elaborate(hir, &config)?)
@@ -428,7 +428,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             File::open(&source)?.read_to_string(&mut file)?;
             let hir = frontend.parse(&file, base_path, &doppio::file_opener)?;
             let journal: doppio::elaboration::Journal =
-                doppio::elaborate(hir, frontend.defaults())?;
+                doppio::elaborate(hir, &frontend.elaboration_defaults())?;
             let compression = if no_compression {
                 doppio::Compression::None
             } else {
--- a/crates/doppio-cli/src/main.rs
+++ b/crates/doppio-cli/src/main.rs
@@ -228,12 +228,19 @@ fn load_proto_journal_with_tolerance(
         let base_path = path.parent().unwrap_or(std::path::Path::new(""));
         let mut file = String::new();
         File::open(path)?.read_to_string(&mut file)?;
-        let mut hir = frontend.parse(&file, base_path, &doppio::file_opener)?;
-        if let Some(fraction) = tolerance_override {
-            hir.global_context.tolerance_mode =
-                doppio::resolution::ToleranceMode::FractionOfSmallestPrecision(fraction);
-        }
-        Ok(doppio::elaborate(hir)?)
+        let hir = frontend.parse(&file, base_path, &doppio::file_opener)?;
+        // Start from the matching tool's default semantics; override the
+        // tolerance fraction if `--tolerance` was passed on the CLI.
+        let config = match tolerance_override {
+            None => frontend.defaults().clone(),
+            Some(fraction) => doppio::resolution::ElaborationConfig {
+                tolerance_mode: doppio::resolution::ToleranceMode::FractionOfSmallestPrecision(
+                    fraction,
+                ),
+                ..frontend.defaults().clone()
+            },
+        };
+        Ok(doppio::elaborate(hir, &config)?)
     }
 }
 
@@ -420,7 +427,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             let mut file = String::new();
             File::open(&source)?.read_to_string(&mut file)?;
             let hir = frontend.parse(&file, base_path, &doppio::file_opener)?;
-            let journal: doppio::elaboration::Journal = doppio::elaborate(hir)?;
+            let journal: doppio::elaboration::Journal =
+                doppio::elaborate(hir, frontend.defaults())?;
             let compression = if no_compression {
                 doppio::Compression::None
             } else {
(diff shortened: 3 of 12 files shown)


alevy:
`default()` is a bad name... default-what? default frontend? default semantic configuration? Also LazyLock is a sin and should be avoided. If the configuration type can't be a const, just construct and return an owned type in the trait method.

Finally, benchmarks failing to compile.

alevy:
Pushed \`7747710\` addressing all three points:

1. **Renamed \`defaults()\` → \`elaboration_defaults()\`**. Agree the unqualified \"defaults\" was vague — \"elaboration defaults\" makes clear what's being defaulted.
2. **Dropped LazyLock**. The trait method now returns owned \`ElaborationConfig\` constructed inline. Each frontend module also exports a free helper (\`ledger_defaults()\` / \`hledger_defaults()\` / \`beancount_defaults()\`) for test code that doesn't want to instantiate a \`Frontend\` value just to pull defaults.
3. **Fixed the benchmark**. \`crates/doppio/benches/pipeline.rs\` still used \`hir.try_into()\` for elaboration; switched to \`doppio::elaborate(hir, &doppio::grammars::ledger::ledger_defaults())\`.

All checks still green locally: 304 lib tests, 14 parity fixtures, 12/12 self-test assertions, fmt + clippy clean, benches build.

PR #282, opened 2026-05-09
fix(elaborator): intent-driven canonical-cost balance check (#281)

### LINKED ISSUE #281: [parser/elaborator] Carry @-price precision intent through HIR for principled balance check
## Context

doppio's ledger frontend currently rejects journals whose `@`-priced transactions don't balance to **exact** zero. ledger-cli accepts the same journals because of a per-posting price-rounding compensation in its balance check. This issue captures a principled doppio-native solution: encode the price's precision intent at parse time, carry it through the HIR, and consume it in the elaborator's balance check — without a tolerance heuristic.

## Empirical anchor: standard.dat:1880

```ledger
2003/06/19 * cbb4cc49824bf79827cde838e005848027ca0a38
    c56a21d2…  331.296869 LMVTX @ $53.6599999999999999998612221219
    c56a21d2…  -523.942988 EEEEE @ $33.9299999999999999998438748872
    c56a21d2…  55.981364 LMVTX @ $53.6599999999999999998612221219
    c56a21d2…  -88.534054 EEEEE @ $33.9299999999999999998438748872
```

Computing each posting's cost at full Decimal precision and summing produces a non-zero residue of `$0.0039477200000000418773963` — the high-precision prices (28 decimal places, indicative of an IEEE-double source) don't cancel exactly. doppio rejects with `TransactionDoesNotBalance`. ledger-cli accepts.

Surfaced while evaluating standard.dat parity (#256). 432+ transactions before this one parse and balance cleanly under doppio.

## What ledger-cli actually does (xact.cc:872-904)

Not a tolerance window. Per-posting price-rounding compensation:

```c++
// for each @-priced posting:
amount_t exact(post->cost->rounded());
const int price_prec = exact.precision() - post->amount.precision();
if (price_prec > 0 && exact.precision() > price_prec)
    rounding_adj += exact.roundto(price_prec) - exact;

// accept iff balance + Σ rounding_adj == 0
```

The motivating example in the same file (line 875-878):

> When a posting has 0.50 bread @ \$3.99, the exact cost \$1.995 is kept at full precision during accumulation. If the user supplies the rounded total \$9.49 instead of the exact \$9.480, we accept the transaction by verifying: balance + Σ(roundedCost - exactCost) == 0.

This is a **journal-format affordance** for human-rounded sums and machine-generated high-precision inputs (the standard.dat case is the latter — IEEE-double prices serialised with all the trailing imprecision intact). ledger separates "displayed precision" from "computational precision" in its amount type and uses the displayed precision as the rounding target.

## Proposed doppio architecture: intent-driven, not tolerance-driven

Capture the user's intended price precision **at parse time** (where it's known authoritatively) and carry it through the HIR to the elaborator. The elaborator's balance check becomes a clean derivation, not a heuristic.

### Step 1: parser captures intent

`ast::ValueExpr::Amount` (or the price-annotation type, depending on where prices land) gains an `intended_scale: u32` derived from the parsed literal. For `$3.99` this is `2`; for `$53.6599999999999999998612221219` this is `28` (whatever the input literally has). `rust_decimal::Decimal` already carries `scale` exactly, so this is data the parser already has — we just need to keep it through the AST layer instead of letting it dissolve into post-multiplication scale.

### Step 2: HIR preserves intent

`resolution::Posting`'s price annotation field carries `intended_scale` alongside the price `Decimal`. (The `Decimal` field stays unchanged — we keep the exact value; intent is metadata.)

### Step 3: elaborator uses intent for canonical cost

For each `@`-priced posting, the elaborator computes:

```rust
let exact_cost = qty * price.value;  // scale = qty.scale + price.value.scale
let canonical_cost = exact_cost.round_dp(qty.scale + price.intended_scale);
```

The transaction balance check is `Σ canonical_cost == 0` per commodity — exact zero, no tolerance. Postings whose `@` price has high actual precision but low intended precision are rounded to the user's stated intent before the balance check; the rounding error is exact and well-defined.

This is mathematically equivalent to ledger's "compute exact, then add rounding adjustment, then check" — but inverted: compute intent-rounded directly, then check. Cleaner data flow.

### Step 4: existing `ToleranceMode::FractionOfSmallestPrecision` stays orthogonal

Beancount-style fractional tolerance (a general "sloppy balance" affordance) is not the same concept and stays as its own mode for journals that want it. Ledger frontend defaults shift from `FractionOfSmallestPrecision(0)` (current, requires exact zero) to "use HIR-level intent-rounded canonical cost." Either as an explicit new `ToleranceMode` variant or as the natural behavior when intent data is present in the HIR.

## Why this is better than a tolerance window

- **No heuristic in the elaborator**: the rounding decision is data carried from the parser, not policy applied at balance check.
- **Exactly explained, not approximately**: the balance check still requires exact zero; we just redefine "cost" to use intent-rounded values. There's no "fudge factor."
- **Encoding matches the actual reality**: prices in the input have a stated precision; we should respect that as the authoritative precision rather than blindly using the post-multiplication scale.
- **Composable**: the same HIR data could inform display rendering, lot-cost printing, and other consumers without each of them re-deriving the heuristic.

## Account-balance impact

For the empirical case: the account `c56a21d2…` receives 4 postings totaling `+387.278233 LMVTX` and `-612.477042 EEEEE`. These are stated quantities at the user's exact scale (6 decimals); they're not rounded. The intent-precision rounding only affects the **implicit `$` balance check** (the cost-basis commodity that doesn't land on any explicit posting in this transaction). It doesn't change any account's tracked balance.

For transactions where an `@`-priced cost DOES land on an account (e.g., a buy where the auto-balanced leg is the cash account): the elaborated Journal stores the exact `Decimal` value as the account's contribution. Display rounding is a separate concern at the CLI/output layer — the elaborated truth is exact.

## Implementation sketch

1. **Parser**: thread `Decimal::scale` from the parsed `@` price literal into the AST node — likely `ast::PriceAnnotation` or wherever `@`-prices live. `rust_decimal::Decimal` already preserves scale, so the data is in hand; the work is making sure the AST carries a separate `intended_scale` field even after parser-internal arithmetic that would otherwise dissolve the per-component scale.
2. **HIR**: mirror the parser's `intended_scale` field on the resolved posting type that carries price annotations.
3. **Elaborator**: at the place where per-posting cost is added to `transaction_state`, compute the intent-rounded canonical cost (`exact.round_dp(qty.scale + intended_scale)`). The balance check stays "exact zero per commodity"; the rounding is applied to the *cost contribution*, not to a tolerance window.
4. **Tests**:
   - standard.dat:1880 4-leg buy/sell elaborates and balances cleanly.
   - The ledger-cli motivating example (`0.50 bread @ \$3.99` with user-supplied `\$9.49` rounded total) elaborates cleanly.
   - A doppio-native journal with exact-precision prices (e.g. `\$3.99` written as scale-2) continues to elaborate as before — no change.
   - Existing `FractionOfSmallestPrecision` tolerance tests pass unchanged (orthogonal).
5. **Defaults**: `ledger_defaults()` and `hledger_defaults()` use the intent-driven mode. Beancount frontend continues with `FractionOfSmallestPrecision` (its convention).

## Acceptance

- `tests/parity/ledger-standard.dat` (vendored under #256) elaborates cleanly under doppio.
- The first 432+ transactions of standard.dat that already pass continue to pass.
- ledger-cli's empirical \$0.004 imbalance at line 1880 is correctly explained by the per-posting intent-rounding adjustments.
- All existing tolerance tests pass without modification.
- Document in `docs/SUPPORTED_FEATURES.md` how doppio's intent-driven canonical-cost differs from ledger-cli's `keep_precision` mechanism — same observable behavior, cleaner data model.

## Out of scope

- Cleaning up Beancount's `FractionOfSmallestPrecision` algorithm (the doc comment in `resolution.rs:489-491` claims it matches ledger-cli, which we now know is wrong; should be reworded to "matches Beancount's `inferred_tolerance_default`").
- Display-rendering decisions (the existing CLI display layer chooses precision; this issue is about elaboration semantics, not output formatting).
- Tolerance modes other than `FractionOfSmallestPrecision` and the new intent-driven one.

## Related

- #256 — standard.dat parity (this issue is the primary blocker for wiring it into POSITIVE)
- #270 — wow.dat parity gap inventory (different concern: chain-canonicalisation display, not balance-check)
- ledger-cli source: `xact.cc:872-904` (balance check) and `xact.cc:195-234` (`set_cost_prec_if_needed`)
- ledger-cli issue #1125 — the original motivation for ledger's price-rounding compensation

alevy:
## Summary

- **Architecture**: parse → elaboration data flow. A pre-elaboration scan (`collect_commodity_scales`) walks all direct posting amounts in the HIR and records the minimum decimal scale per commodity. At cost-computation time, each `qty * price` product for `@`-priced postings is rounded to the price commodity's natural scale before being accumulated into `transaction_state`. The balance check continues to require exact zero.

- **Empirical anchor — standard.dat:1880 before/after**:
  ```
  2003/06/19 * cbb4cc49824bf79827cde838e005848027ca0a38
      c56a21d2…  331.296869 LMVTX @ $53.6599999999999999998612221219
      c56a21d2…  -523.942988 EEEEE @ $33.9299999999999999998438748872
      c56a21d2…  55.981364 LMVTX @ $53.6599999999999999998612221219
      c56a21d2…  -88.534054 EEEEE @ $33.9299999999999999998438748872
  ```
  Before: `TransactionDoesNotBalance(Amount({"$": 0.0039477200000000418773963}))`. After: elaborates cleanly. The `$0.004` residue is fully explained by the 28-decimal-place prices being IEEE-double serialisations of 2-decimal-place values; prior transactions in standard.dat (`$474.31`, `$14.91`, …) establish a natural `$`-scale of 2, and `round_dp(2)` on each per-posting cost makes the four costs sum to exactly zero.

- **ledger-cli motivating example**: `0.50 bread @ $3.99` × 2 postings with `$-4.00` cash. The `$-4.00` direct amount establishes `$`-scale 2; `0.50 × 3.99 = 1.995` rounded to 2dp = `$2.00`; two postings sum to `$4.00`, matching `$-4.00`. Accepted.

- **Implementation is NOT a tolerance window**: no `ToleranceMode` change. The balance check still requires exact zero; "cost" is redefined to the intent-rounded value. This matches ledger-cli's `xact.cc:872-904` semantics but is data-driven from a pre-pass rather than a heuristic applied at balance-check time.

## Test plan

- [x] `test_intent_rounded_cost_standard_dat_1880_repro` — four-leg LMVTX/EEEEE transaction with 28-decimal prices elaborates cleanly; LMVTX and EEEEE quantities preserved exactly
- [x] `test_intent_rounded_cost_ledger_motivating_example` — `0.50 bread @ $3.99` × 2 with `$-4.00` cash elaborates
- [x] `test_intent_rounded_cost_exact_precision_unaffected` — `1.00 cup @ $3.99` with `$-3.99` cash; no regression for clean-precision prices
- [x] `test_intent_rounded_cost_high_precision_imbalance_rejected` — `1 G + -1.001 G` (no `@` price, genuine imbalance) still rejected
- [x] All 405 pre-existing lib tests pass without expectation changes
- [x] All 49 parity integration tests pass
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test --doc` clean
- [x] `python3 scripts/parity_check.py --negative` (4/4 pass)

Closes #281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@
 
 ### Fixed
 
+- **Intent-driven canonical-cost balance check for `@`-priced transactions** (#281):
+  doppio previously rejected transactions whose `@`-price costs did not sum to exactly zero at
+  full `rust_decimal` precision, even when the residue was fully explained by the prices'
+  intended decimal precision. ledger-cli accepts such transactions by rounding each per-posting
+  cost to the price commodity's "display precision" before the balance check.
+
+  The fix: a pre-elaboration scan collects the minimum decimal scale observed for each commodity
+  across all direct posting amounts in the journal. At cost-computation time, each `qty * price`
+  product is rounded to that commodity's natural scale before being accumulated into the
+  transaction balance. The balance check still requires exact zero — "cost" is redefined to use
+  intent-rounded values rather than full-precision products. No tolerance window is introduced.
+
+  This unblocks transactions with high-precision prices that are IEEE-double serialisations of
+  two-decimal-place values (e.g. `$53.6599999999999999998612221219` for `$53.66`): once
+  rounded to the 2-decimal-place natural precision of the `$` commodity established by earlier
+  transactions, the four-leg buy/sell in `standard.dat:1880` sums to exactly zero.
+
+  **Behaviour change**: journals where `@`-price products do not sum to exact zero at full
+  `rust_decimal` precision will now be accepted rather than rejected, provided the residue is
+  explained by per-commodity precision. Journals where direct posting amounts are written as
+  whole numbers (e.g. `$100` with scale 0) will have their `@`-price costs rounded to whole
+  numbers; use `$100.00` to avoid this.
+
 - **Auto-balance posting preserves per-lot identity — principled rule, no inventory-account guard** (#276):
   When a transaction's explicit postings carry `{cost}` lot annotations on item commodities (typical of
   inter-account inventory transfers and first-time grants), doppio previously emitted a single aggregated
--- a/crates/doppio/src/elaborator.rs
+++ b/crates/doppio/src/elaborator.rs
@@ -583,6 +583,12 @@ pub fn elaborate(
         let assertion_scope = config.assertion_scope;
         let lot_validation_mode = config.lot_validation_mode;
 
+        // Pre-scan: collect the minimum decimal scale seen for each commodity
+        // across all direct posting amounts in the journal. This mirrors
+        // ledger-cli's "commodity display precision" tracking and is used by
+        // `at_price_cash` to round intent-precision costs (see #281).
+        let commodity_scales = collect_commodity_scales(&value);
+
         // Auto-rule queries are pre-compiled to regexes during resolution;
         // here we just need to extract the rules so we can iterate
         // `value.entries` (consuming it) without conflicting with a borrow
@@ -1047,12 +1053,42 @@ pub fn elaborate(
                                                 Some((v, c))
                                             }
                                             ast::LotPricing::Unit(expr) => {
+                                                // Intent-driven canonical cost (#281):
+                                                // round `qty * price` to the price
+                                                // commodity's natural decimal precision
+                                                // before accumulating into the transaction
+                                                // balance. This matches ledger-cli's
+                                                // per-posting price-rounding compensation
+                                                // (xact.cc:872-904) without a tolerance
+                                                // window — the balance check still requires
+                                                // exact zero, but "cost" is redefined as
+                                                // the intent-rounded value rather than the
+                                                // full-precision product.
+                                                //
+                                                // `commodity_scales` (pre-computed before
+                                                // the main elaboration loop) holds the
+                                                // minimum scale seen for each commodity
+                                                // across all direct posting amounts in the
+                                                // file. For `$`, prior `$474.31` amounts
+                                                // establish a scale of 2. High-precision
+                                                // prices like `$53.659999...28 digits` are
+                                                // then rounded to 2 decimal places (cents)
+                                                // rather than 27, removing the IEEE-double
+                                                // noise and producing a zero-sum balance.
                                                 let (v, c) = evaluator::eval_and_normalize_amount(
                                                     expr,
                                                     entry_context,
                                                     state,
                                                 )?;
-                                                Some((v * value, c))
+                                                let exact_cost = v * value;
+                                                let canonical_cost = if let Some(&scale) =
+                                                    commodity_scales.get(&c)
+                                                {
+                                                    exact_cost.round_dp(scale)
+                                                } else {
+                                                    exact_cost
+                                                };
+                                                Some((canonical_cost, c))
                                             }
                                         })
                                     };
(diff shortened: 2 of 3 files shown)


alevy:
Closing in favor of a resolution-layer redo.

The implementation here is empirically correct (5 tests pass, CI green, standard.dat:1880 elaborates cleanly), but the architecture places the commodity-scale inference in elaborator.rs. That's a layering violation: the inference is derivable from parsed source, so it belongs in resolution and the result should live on HIR (likely \`CommodityProperties\`). Putting it in elaborator forces a uniform rule across frontends and walks the AST a second time for no benefit.

The redo will move \`collect_commodity_scales\` to resolution as a side-effect of the existing walk, store the per-commodity scale on \`CommodityProperties\` in HIR, and have elaborator read from the HIR rather than re-derive. Same observable behavior, cleaner layering, and per-frontend customisation becomes possible later without elaborator changes.

Issue #216, opened 2026-05-07
[cli] richer output formatting: tree-rollup balance, color, terminal-width adaptation


alevy:
## Background

\`dop\`'s output today is correct but spartan. This issue tracks improvements that make the CLI more pleasant for interactive use, without changing any underlying semantics or wire format. Low priority — file new sub-issues if any individual item starts to need its own thread.

### Hierarchical balance rollup

\`dop balance\` already defaults to tree mode, but the current implementation is **indented-flat**: each row prints the literal account's direct balance with depth-based indentation; the parent line is empty when the parent has no direct postings. ledger-cli's \`bal\` *rolls up* — the parent row shows the subtree sum.

Concretely, given:

\`\`\`
Assets:Bank:Checking  +100 USD
Assets:Bank:Savings   +200 USD
Assets:Bank            0       (no direct postings)
\`\`\`

doppio prints:

\`\`\`
                                  Bank
        100 USD                     Checking
        200 USD                     Savings
\`\`\`

ledger-cli prints:

\`\`\`
        300 USD            Assets:Bank
        100 USD              Bank:Checking
        200 USD              Bank:Savings
\`\`\`

The infrastructure for the rollup landed with PR #215: \`subtree_commodity_balance\` in \`crates/doppio/src/elaborator.rs\` is exactly the rollup primitive. The CLI just needs to call through it for each rendered row.

Also covered:
- Refactor the ad-hoc colon walking in \`crates/doppio-cli/src/main.rs:763-820\` (\`truncate_account\`, \`rsplit_once(':')\`, \`chars().filter(|c| c == ':').count()\`) onto the same shared helper so colon-segment interpretation lives in one place.

### Colored terminal output

When stdout is a TTY, color-code:
- Account types (Assets / Liabilities / Equity / Income / Expenses) per ledger-cli convention.
- Negative vs positive amounts.
- The cleared-state marker (\`*\` / \`!\`) on \`register\` output.

Off by default, on with \`--color=always\` / \`--color=auto\`. Use the existing crate ecosystem (\`anstream\` or similar).

### Terminal-width-aware register output

\`dop register\` currently uses fixed-width columns. ledger-cli adapts the description / payee column to the available terminal width and truncates / wraps long lines. Worth doing.

### Other small things, file as sub-issues if they grow

- Right-align amounts on the column boundary so multi-commodity totals line up vertically.
- \`--currency-symbol-position=before|after\` toggle.
[...]

PR #307, opened 2026-05-11
feat(cli): grand-total row + -X/--exchange for dop balance (refs #216)

alevy:
## Summary

- Adds a 20-dash separator and grand-total row after the last account row in `dop balance` text output, matching ledger-cli's `bal` footer format.
- Adds `--color={auto,always,never}` as a global flag; negative grand-total amounts are rendered red when color is enabled.
- The `-X`/`--exchange` flag was already wired in the codebase; this PR adds missing-rate-fallback tests and partial-conversion (mixed-total) tests to cover the edge cases specified in #216.

## Before / After

**Before:**
```
               $ -50  Assets:Checking
            EUR -100
                $ 50  Expenses:Food
             EUR 100  Expenses:Travel
```

**After (`dop balance --flat`):**
```
               $ -50  Assets:Checking
            EUR -100
                $ 50  Expenses:Food
             EUR 100  Expenses:Travel
--------------------
                 $ 0
               EUR 0
```

**With `-X $`:**
```
           $ -160.00  Assets:Checking
                $ 50  Expenses:Food
            $ 110.00  Expenses:Travel
--------------------
              $ 0.00
```

## -X semantic: approach (b) aggregate-then-convert

The existing `-X`/`--exchange` implementation uses **approach (b)**: amounts are aggregated per (account, native-commodity) first, then the per-account balance totals are converted using the most-recent `P` rate on or before `--end` (or the latest rate if `--end` is not set).

Approach (a) (per-posting-date conversion before aggregation) would require tracking the transaction date through the balance aggregation loop — currently only the posting amount and account are tracked. Since the existing implementation was already in place and the aggregation pipeline does not expose per-posting dates at the point of balance accumulation, approach (b) was retained. This is documented in help text.

## Missing-rate fallback

When no `P` rate exists for a (source, target) pair:
- The amount is left in its native commodity (no error, no drop).
- A warning is printed to stderr: `warning: no FX path from GBP to USD; leaving 100 GBP unconverted`.
- The grand-total row includes one line per residual native commodity alongside any successfully-converted amounts.

This was empirically tested via the `balance_exchange_missing_rate_leaves_amount_native` and `balance_exchange_partial_conversion_mixed_total` tests.

## Total row scope

- **Tree mode**: total aggregates top-level account subtrees only (depth=1 accounts). Summing over all depths would double-count intermediate nodes (e.g. `Assets`, `Assets:Bank`, `Assets:Bank:Checking` would count the same posting three times).
- **Flat mode**: total is the sum of all displayed rows.
- When no accounts match the filter (e.g. `--tag nonexistent`), no separator or total row is emitted — matching ledger-cli's behaviour.

Refs #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Grand-total row for `dop balance`** (refs #216): after the last account row,
+  `dop balance` now emits a 20-dash separator (`--------------------`) followed by
+  a grand-total row (one line per commodity). In tree mode the total is the sum of
+  all top-level account subtrees; in flat mode it is the sum of all displayed rows.
+  Negative totals are colored red when `--color=always` or when stdout is a TTY
+  with `--color=auto`. The separator and total are suppressed when no accounts match
+  the filter, matching ledger-cli's behaviour.
+
+- **`-X` / `--exchange` flag for `dop balance`** (refs #216): converts all commodity
+  balances to a target commodity using `P` price directives. When no rate is available
+  for a (source, target) commodity pair the original amount is left unconverted and a
+  warning is printed to stderr. Uses the most-recent `P` directive on or before the
+  `--end` date (or the latest available if `--end` is not set). Matches ledger-cli's
+  `-X`/`--exchange` semantics.
+
+- **`--color={auto,always,never}` global flag**: controls ANSI color output. `auto`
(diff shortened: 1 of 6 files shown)


alevy:
Closing — the worktree this PR was built in was branched from a commit predating #305 and #306 merging, so the diff against current main effectively deletes those PRs' work. Re-dispatching a fresh PR that adds the grand-total row + missing-rate-fallback tests on top of current main without touching the color or writer modules.

Issue #319, opened 2026-05-16
[categorize] cold-start: classifier returns empty for unseen known_account values


alevy:
## Summary

`Index::suggest` short-circuits to an empty result when `query.known_account` doesn't exist in `Index.by_known` (`crates/doppio-categorize/src/query.rs:119`). The classifier has no way to generalize across accounts that share an account-tree prefix, so onboarding a new credit card or routing direct-deposit salary to a new checking account is a total cold start, even when the user has years of history on sibling accounts.

## Motivating examples

- User adds `Liabilities:Visa:NewCard`. They have a thousand `Liabilities:Visa:OldCard` samples (closed card, replaced). Every import on the new card returns zero suggestions until they manually categorize a critical mass.
- User opens `Assets:Checking:NewBank`. Direct-deposit salary that used to route through `Assets:Checking:OldBank → Income:Wages:Day-Job` is now uncategorizable on import.

The signal needed to predict these correctly is present in the journal — it's just behind a hash-table lookup that requires an exact account-name match.

## How this differs from related issues

- **#317** (cross-known-account matching for shared-merchant cases) — same merchant booked across *multiple already-populated buckets*; the proposed fix aggregates pre-existing sample pools. This issue is about a new bucket that doesn't yet exist: siblings have history, the queried bucket itself is empty.
- **#316** (alias resolution) — the bucket exists but is keyed under the elaborated post-alias name. Orthogonal. Alias resolution happens first; hierarchical fallback fires after, only if the post-alias bucket is still empty or sparse.

## Proposed design

A candidate-selection stage that activates when the per-known-account bucket yields no candidates (after both exact-match and token-IDF in the bucket-local stages). It expands the candidate pool to **prefix-sibling buckets**, ranked by account-tree distance, with a multiplicative weight penalty per level of generalization.

Account-prefix decomposition follows the doppio convention (`:` separator):

- `Liabilities:Visa:NewCard` walks up to tiers `Liabilities:Visa:NewCard`, `Liabilities:Visa`, `Liabilities`.
- For each tier above the leaf, collect buckets in `Index.by_known` whose account name starts with `tier:` or equals `tier`.
- Run the configured strategy (exact-match → token-IDF) against the combined sample list, multiplying each sample's score by a tier-distance weight (e.g. `1.0` same-bucket, `0.5` same-parent, `0.25` same-grandparent, `0.1` same-top-level).

The sign filter and amount-similarity weight (`src/query.rs:175-188`) carry over unchanged.

[...]

PR #321, opened 2026-05-16
[categorize] ScoringStrategy::HierarchicalHybrid for cold-account recovery

alevy:
## Summary

- Adds `ScoringStrategy::HierarchicalHybrid` to `doppio-categorize`: when the per-known-account bucket yields no candidates, walks up the account tree (colon-separated tiers) and pools prefix-sibling buckets, applying a multiplicative weight penalty per level of generalization.
- Wires `--strategy hierarchical` (also accepts `hierarchical-hybrid`) and `--prefix-weights W1,W2,...` (default `1.0,0.5,0.25,0.1`) into `eval_holdout`.
- Adds 10 unit tests covering prefix walk, tier fallback, sibling exclusion, weight depth stopping, and edge cases (empty weights, single-tier accounts).
- Adds 2 integration tests reproducing the "new credit card" scenario.

Part of #319.

## Algorithm

Tier 0: run `Hybrid` (exact-match → token-IDF) against the exact bucket, multiplied by `prefix_weights[0]`. If candidates found, return immediately.

Tiers 1+: walk up the account tree. At each tier `k`, collect every bucket whose account name equals `prefix_k` or starts with `prefix_k:`, excluding the original bucket. Pool their samples into a virtual bucket, run Hybrid scoring, multiply by `prefix_weights[k]`. Return at the first tier with candidates.

## Edge case decisions

| Case | Behavior |
|---|---|
| No `:` in account name | Tier 0 only; no fallback walk |
| `prefix_weights` shorter than tree depth | Stop walking when weights are exhausted |
| Empty `prefix_weights` | Normalized to `[1.0]`; tier-0 only (equivalent to plain Hybrid) |
| Original bucket matches its own prefix | Excluded from sibling pool to prevent double-counting |
| Score scaling vs. confidence normalization | Tier weight multiplies raw match scores before `rank_candidates`; `rank_candidates` divides by `total_weight` as usual, so tier weights affect only which tier is chosen (first non-empty tier wins), not absolute confidence within a tier |

### Leave-One-Account-Out (min 5 txns, `--prefix-weights 1.0,0.5,0.25,0.1`)

| Corpus | Strategy | Cohort top-1 | Cohort top-3 | Aggregate top-1 | Aggregate top-3 |
|---|---|---|---|---|---|
| Household | hybrid (baseline) | 0.1% | 0.6% | 0.1% | 0.2% |
| Household | hierarchical (this PR) | **34.6%** | **41.3%** | **9.5%** | **12.2%** |
| BB | hybrid (baseline) | 0.0% | 0.0% | 0.0% | 0.0% |
| BB | hierarchical (this PR) | 0.0% | 0.0% | 0.0% | 0.0% |

Household lift: +34.5pp cohort top-1. BB remains 0% — the two eligible accounts (`Assets:Checking:Mercury:7920` and `Liabilities:Credit Card`) have no prefix siblings in the index, so the fallback walk finds nothing regardless of strategy.

### Uniform holdout non-regression

| Corpus | Strategy | Top-1 | Top-3 |
|---|---|---|---|
| Household | hybrid | 73.3% | 79.7% |
| Household | hierarchical | 73.3% | 80.1% |
| BB | hybrid | 77.8% | 77.8% |
| BB | hierarchical | 77.8% | 77.8% |

No regression on uniform holdout (tier-0 is identical to Hybrid behavior, so existing populated accounts are unaffected).

## Test plan

- [x] `cargo fmt` — clean
- [x] `cargo test -p doppio-categorize` — 38 unit + 14 integration tests pass
- [ ] CI green

### DIFF
--- a/crates/doppio-categorize/examples/eval_holdout.rs
+++ b/crates/doppio-categorize/examples/eval_holdout.rs
@@ -21,8 +21,9 @@
 //!   <journal-path> \
 //!   [--import-regex REGEX] \
 //!   [--no-amount-weighting] \
-//!   [--strategy exact|token-idf|hybrid] \
+//!   [--strategy exact|token-idf|hybrid|hierarchical] \
 //!   [--df-threshold N] \
+//!   [--prefix-weights 1.0,0.5,0.25,0.1] \
 //!   [--holdout 0.10] [--seed N] \
 //!   [--cold-account ACCOUNT] \
 //!   [--leave-one-account-out [MIN_TXN_COUNT]]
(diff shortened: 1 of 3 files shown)


alevy:
Closing in favor of the v0.2 reframe in #319.

A follow-up probe measured what happens if we drop `known_account` from candidate selection entirely. Result: household LOAO cohort top-1 **44.0%** (vs 34.6% here), aggregate **18.7%** (vs 9.5%), no uniform-holdout regression. The simpler design strictly dominates this PR's hierarchical fallback on every measured metric.

The hierarchical walk could only recover signal that happened to share an account-tree prefix. The actual dominant signal — `payee → counter_account` — crosses account-tree boundaries freely (same merchant paid from checking and credit), and a prefix walk can't see those samples. BB's persistent 0% LOAO with this PR was the smoking gun.

Shipping this would entrench an intermediate architecture we'd have to deprecate immediately. The eval-mode infrastructure (#320) and the cold-start framing both remain — only the proposed *fix* changes. See #319's latest comment for the new direction.

PR #330, opened 2026-05-17
[doppio] add HirBuilder for external HIR construction

### LINKED ISSUE #328: [doppio] external HIR construction is blocked: pub(crate) entries + ResolutionEntry
## Summary

`Frontend::write_journal(&HIR, &mut Writer)` is the documented way to serialise a journal in any of the three frontend formats. The library's public surface implies that downstream consumers can use this to write transactions in their preferred frontend syntax. In practice, external callers can't construct an `HIR` to feed into it, because two layers of the type are crate-private.

## Concrete blocker

In `crates/doppio/src/resolution.rs`:

```rust
pub struct HIR {
    pub(crate) entries: Vec<ResolutionEntry>,   // ← can't write to from outside
    pub contexts: Vec<Context>,
    pub global_context: GlobalContext,
    pub prices: Vec<HistoricalPrice>,
    pub(crate) auto_rules: Vec<ResolvedAutoRule>, // ← can't write to from outside
}

pub(crate) struct ResolutionEntry { … }   // ← can't construct from outside either
```

`HIR::default()` works (gives an empty HIR), but `entries` is `pub(crate)` so external code can't append transactions. Even if `entries` were public, `ResolutionEntry` is itself `pub(crate)`, so the element type can't be constructed.

`resolution::Transaction` and `resolution::Posting` *are* fully public structs with all-public fields, so the leaf types are usable. The wrapping container is what's blocked.

## Use cases blocked

- **Tools that emit new transactions**: a personal-finance app that wants to construct synthetic transactions (from CSV imports, from auto-categorize suggestions, etc.) and serialise them into the user's chosen frontend syntax. Today it can't use `write_journal`; it has to hand-roll a stringifier per format.
- **Programmatic fixture generation**: tests / examples that want to build a journal in code and write it in a specific format.
- **Round-trip transformations from external crates**: parse → modify → write, where the modification adds new transactions rather than just mutating existing ones (most existing mutations to fully-public fields like `Transaction.metadata` already work).

## Possible fixes

Several, with different trade-offs:

1. **Make `entries` and `auto_rules` `pub` + make `ResolutionEntry` public.** Minimal change. Cost: locks in the current internal shape (no longer pub(crate) means breaking changes if internals evolve). Forces every caller to know about `Context` indices, `ResolutionEntry` discriminants, etc.

2. **Add an `HirBuilder` type** that wraps construction: `HirBuilder::new().push_transaction(t).push_price(p).build()`. Hides the entry-list-with-context-indices structure. Lets internals stay private and evolve.

3. **Add a simpler write API that bypasses HIR for the common case.** `Frontend::write_transactions(&[Transaction], &mut Writer)` would let callers serialise transactions without constructing the HIR scaffolding around them. Doesn't cover writing alias/commodity directives — but most "I want to emit transactions" use cases don't need those.

(1) + (3) probably together: (3) covers the common case cleanly, (1) unblocks the full-fidelity case.

## Out of scope

- Constructing an `elaboration::Journal` from scratch. Different layer, different problem. Elaboration is the post-balanced read model used by e.g. `doppio-categorize`; it doesn't round-trip back to source text.
- A general "edit a parsed HIR" API beyond what's needed to add transactions/prices/etc. Mutation of existing entries is mostly already possible via the public fields on `Transaction` and `Posting`.

## Acceptance

A downstream crate can:

1. Construct a `resolution::Transaction` with at least one `resolution::Posting`.
2. Wrap it into something writable.
3. Call `LedgerFrontend.write_journal(...)` (or hledger / beancount equivalents) and get back valid source text.

All without needing `pub use` of a `pub(crate)` type from inside the doppio crate.

Demonstration test should live in `crates/doppio/tests/` exercising the path from an externally-constructed transaction to written text in each of the three frontends.

alevy:
## Summary

Closes #328.

External crates cannot construct an `HIR` to pass to `Frontend::write_journal` because `HIR.entries` is `pub(crate)` and `ResolutionEntry` / `Entry` are also `pub(crate)`. This blocks the documented use case of building synthetic transactions and serialising them.

### Design note

Three options were considered:

1. **Flip `pub(crate)` → `pub`** on `HIR.entries` and the entry types. Rejected: forces external callers to understand `context_id` (a resolution-stage artifact that's meaningless to them) and locks in the `ResolutionEntry` shape as a public API.

2. **`HirBuilder`** (this PR): a builder that accepts `Transaction` and `HistoricalPrice` values and produces a write-ready `HIR`, with `context_id: 0` on every entry and an empty-but-valid contexts table. Writers iterate `entries` and `prices` only — they never dereference `contexts` — so the empty table is safe for the write path.

3. **`Frontend::write_transactions` shortcut** that bypasses HIR entirely. Rejected: duplication of formatting logic, and doesn't cover the read+modify+write flow.

### Public API delta

```rust
// crates/doppio/src/resolution.rs
pub struct HirBuilder { /* private */ }

impl HirBuilder {
    pub fn new() -> Self;
    pub fn push_transaction(self, txn: Transaction) -> Self;
    pub fn push_price(self, price: HistoricalPrice) -> Self;
    pub fn build(self) -> HIR;
}

impl Default for HirBuilder { ... }
```

`HirBuilder` is accessible at `doppio::resolution::HirBuilder` (the `resolution` module is already `pub`).

### Round-trip tests

`crates/doppio/tests/hir_builder.rs` adds two integration tests that exercise only the public API:

- **`hir_builder_writes_through_all_frontends`**: constructs a 2-posting transaction, builds the HIR, writes through `LedgerFrontend`, `HledgerFrontend`, and `BeancountFrontend`, and re-parses each output through the same frontend — confirming the written text is valid in all three formats.
- **`hir_builder_handles_prices`**: same pattern with a `HistoricalPrice` pushed alongside the transaction; asserts that each frontend emits the expected price directive (`P` for ledger/hledger, `price` for beancount).

### Out of scope

- `push_assertion` / `push_pad`: not needed for the named use case. A comment in the builder docs notes they could be added if a consumer asks.
- Elaboration compatibility: `HirBuilder`-produced HIRs are intentionally write-path-only. The doc warns that passing them to `elaborate()` may panic on the empty contexts table.

## Test plan

- [x] `cargo fmt --workspace` — clean
- [x] `cargo test -p doppio` — 431 tests pass (429 existing + 2 new)
- [x] `cargo test --doc -p doppio` — doc-example runs and passes

### DIFF
--- a/crates/doppio/src/resolution.rs
+++ b/crates/doppio/src/resolution.rs
@@ -1019,6 +1019,92 @@ impl HIR {
     }
 }
 
+/// Build an [`HIR`] from externally-constructed transactions and prices.
+///
+/// [`HIR`] is the canonical input to [`crate::frontend::Frontend::write_journal`].
+/// Its internal shape (per-entry `context_id` indices, alias/commodity context
+/// table) is a resolution-stage artifact that downstream writers don't consume
+/// but that external callers can't populate sensibly. `HirBuilder` hides those
+/// internals and exposes a small, typed surface for the use case of "I have
+/// transactions in memory; serialise them as ledger / hledger / beancount source
+/// text."
+///
+/// HIRs built with `HirBuilder` are intended for the write path only. Passing
+/// a builder-produced HIR to [`crate::elaborate`] may panic because the
+/// contexts table is empty and the elaborator dereferences it by index.
+///
+/// `push_assertion` and `push_pad` are not currently provided; open an issue if
+/// your use case requires them.
+///
+/// # Example
+///
+/// ```rust
+/// use doppio::resolution::{HirBuilder, Transaction, Posting};
+/// use doppio::frontend::Frontend as _;
+/// use doppio::LedgerFrontend;
+/// use chrono::NaiveDate;
+/// use rust_decimal::Decimal;
+///
+/// let txn = Transaction::new(
+///     NaiveDate::from_ymd_opt(2024, 1, 15).unwrap(),
+///     "Groceries",
+/// )
+/// .with_posting(
+///     Posting::new("Expenses:Food").with_amount((Decimal::from(50u32), "USD")),
+/// )
+/// .with_posting(Posting::new("Assets:Checking"));
+///
+/// let hir = HirBuilder::new().push_transaction(txn).build();
+///
+/// let mut out = Vec::new();
+/// LedgerFrontend.write_journal(&hir, &mut out).unwrap();
+/// let text = String::from_utf8(out).unwrap();
+/// assert!(text.contains("Groceries"));
+/// ```
+#[derive(Debug, Default)]
+pub struct HirBuilder {
+    entries: Vec<ResolutionEntry>,
+    prices: Vec<HistoricalPrice>,
+}
+
+impl HirBuilder {
+    /// Creates a new, empty builder.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Appends a transaction to the builder (builder pattern).
+    pub fn push_transaction(mut self, txn: Transaction) -> Self {
+        self.entries.push(ResolutionEntry {
+            context_id: 0,
+            data: Entry::Transaction(txn),
+        });
+        self
+    }
+
+    /// Appends a historical price directive to the builder (builder pattern).
+    pub fn push_price(mut self, price: HistoricalPrice) -> Self {
+        self.prices.push(price);
+        self
+    }
+
+    /// Consumes the builder and returns a finished [`HIR`].
+    ///
+    /// The returned HIR has an empty contexts table (sufficient for the write
+    /// path) and no auto-rules.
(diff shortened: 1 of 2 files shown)


alevy:
Closing in favor of a simpler design (per design discussion).

The builder pattern adds API surface for limited benefit. The realistic consumer shape (e.g. bookie iterating over import sources and appending each row as a transaction) is imperative, not chain-style. A direct `&mut self` method on `HIR` fits the use case better and is smaller.

New design, opening shortly:

- `impl HIR { pub fn new() -> Self; pub fn append_entry(&mut self, entry: Entry); }`
- Visibility flips: `Entry`, `AssertionDirective`, `PadDirective` → `pub`
- `ResolutionEntry` stays private — `append_entry` wraps with the correct `context_id` so callers never have to know about it
- Existing round-trip tests carry over against the new API

The HirBuilder approach can be revisited later if a chain-style use case emerges.

